PR #5, opened 2021-06-03
Add Endpoint For Status Check

sohak22:
Create A Status Endpoint
https://www.notion.so/uwblueprintexecs/Create-a-status-endpoint-eec29cc4976e48e2bfb010e50887d65d

We want to create a **/status** endpoint that returns information that the service is running. 

We want it to return 200 response code and the following message:

```json
{"status": "Healthy!"}

```

Steps to test

1. Run the application in local host : http://localhost:3000/api/status 
2. Page should return healthy status

### DIFF
--- a/pages/api/status.tsx
+++ b/pages/api/status.tsx
@@ -1,0 +1,7 @@
+// Next.js API route support: https://nextjs.org/docs/api-routes/introduction
+import { NextApiRequest, NextApiResponse } from "next";
+
+export default (_req: NextApiRequest, res: NextApiResponse): void => {
+    res.status(200).json({ status: "Healthy" });
+    return;
+};
--- a/pages/api/status.tsx
+++ b/pages/api/status.tsx
@@ -1,6 +1,6 @@
-// Next.js API route support: https://nextjs.org/docs/api-routes/introduction
 import { NextApiRequest, NextApiResponse } from "next";
 
+// Creates status endpoint to check if server is responsive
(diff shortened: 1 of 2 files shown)


pages/api/status.tsx
```diff
@@ -0,0 +1,7 @@
+// Next.js API route support: https://nextjs.org/docs/api-routes/introduction
```
feng-eric: I think you can remove this line

mahadzaryab1: don't need this

pages/api/status.tsx
```diff
@@ -0,0 +1,7 @@
...
+import { NextApiRequest, NextApiResponse } from "next";
+
+export default (_req: NextApiRequest, res: NextApiResponse): void => {
```
mahadzaryab1, line 4: add comment above function explaining what the function does

mahadzaryab1 (changes requested):
Added a couple of comments - some other things to consider:
- You can delete the /hello endpoint as it's replaced by this one 
- PR name should be imperative and descriptive. For example: **_Add Endpoint For Status Check_**

feng-eric (changes requested):

feng-eric (approved):
LGTM!

mahadzaryab1 (changes requested):
I added some comments in my last review - could you take a look at those?

mahadzaryab1 (approved):
LGTM!

PR #6, opened 2021-06-04
Setup Authentication and Boilerplate Sign Up Page

mahadzaryab1:
### Notion ticket link

[Set up Authentication](https://www.notion.so/uwblueprintexecs/Set-up-Authentication-217f2c263d69405fa82fc3975fe15686)

- Added dependencies for `nextauthjs`
- Set up boilerplate authentication flow 
- Did not add any custom callbacks yet as that will need to be done after the schemas have been added 
- Did not create a custom email template - to be done in a separate ticket 
- Did not create a verify page - to be done in a separate ticket

### Steps to test

1. Spin up the app and try entering in your email to ensure that the login flow is working as expected. Note that you will need to verify your email in AWS as we are still in sandbox mode.

-

### Checklist

- [x] My PR name is descriptive and in imperative tense
- [x] I have run the linter
- [x] I have requested a review from the PL, and/or other devs who have background knowledge on this PR or who will be building on top of this PR


pages/api/auth/[...nextauth].js
```diff
@@ -0,0 +1,50 @@
...
+        // use JSON web tokens for session handling
+        jwt: true,
+        // maximum age of an idle session - 30 days
+        maxAge: 30 * 24 * 60 * 60,
+    },
+    jwt: {
+        // JWT secret used for generating a key
+        secret: process.env.JWT_SECRET,
+    },
+    // TODO: define custom callback functions
```
Comment (feng-eric) [inline pages/api/auth/[...nextauth].js:47]: Do we also have to pass database config too? Will that be done at a later time?

pages/api/auth/[...nextauth].js
```diff
@@ -0,0 +1,50 @@
...
+    },
+    session: {
+        // use JSON web tokens for session handling
+        jwt: true,
+        // maximum age of an idle session - 30 days
+        maxAge: 30 * 24 * 60 * 60,
+    },
+    jwt: {
+        // JWT secret used for generating a key
+        secret: process.env.JWT_SECRET,
```
Comment (feng-eric) [inline pages/api/auth/[...nextauth].js:45]: For any environment variables, make sure to put the variable name in `.env.sample`.

prisma/migrations/20210529200211_add_tables_for_nextjs_auth/migration.sql
```diff
@@ -0,0 +1,77 @@
+-- CreateTable
```
feng-eric: What's the reason we are doing a migration for this rather than have a `schema.sql` file in the `prisma` directory?

prisma/migrations/20210529200211_add_tables_for_nextjs_auth/migration.sql
```diff
@@ -0,0 +1,77 @@
...
+CREATE TABLE "Account" (
+    "id" TEXT NOT NULL,
```
feng-eric: What's the reason for the type being TEXT instead of SERIAL for id? Also, would it be better to just define the primary key here? `id SERIAL PRIMARY KEY NOT NULL`

prisma/migrations/20210529200211_add_tables_for_nextjs_auth/migration.sql
```diff
@@ -0,0 +1,77 @@
...
+    "userId" TEXT NOT NULL,
+    "providerType" TEXT NOT NULL,
+    "providerId" TEXT NOT NULL,
+    "providerAccountId" TEXT NOT NULL,
+    "refreshToken" TEXT,
+    "accessToken" TEXT,
+    "accessTokenExpires" TIMESTAMP(3),
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
feng-eric: For anything time stamp, I think we should use TIMESTAMPTZ https://medium.com/building-the-system/how-to-store-dates-and-times-in-postgresql-269bda8d6403

prisma/migrations/20210529200211_add_tables_for_nextjs_auth/migration.sql
```diff
@@ -0,0 +1,77 @@
+-- CreateTable
+CREATE TABLE "Account" (
+    "id" TEXT NOT NULL,
+    "userId" TEXT NOT NULL,
+    "providerType" TEXT NOT NULL,
+    "providerId" TEXT NOT NULL,
+    "providerAccountId" TEXT NOT NULL,
+    "refreshToken" TEXT,
```
feng-eric: For any TEXT fields, there doesn't seem to be too much of a difference between TEXT and VARCHAR in postgres. The only diff is that in VARCHAR we can limit the size. Do you think this additional safety of having a max limit for varchar would be beneficial?

prisma/migrations/20210529200211_add_tables_for_nextjs_auth/migration.sql
```diff
@@ -0,0 +1,77 @@
...
+    "accessToken" TEXT,
+    "accessTokenExpires" TIMESTAMP(3),
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updatedAt" TIMESTAMP(3) NOT NULL,
+
+    PRIMARY KEY ("id")
+);
+
+-- CreateTable
```
feng-eric: For comments, I think having the same comment isn't too useful since the actual line already says CREATE TABLE. I think it would be beneficial to say the specific table you're creating

prisma/migrations/20210529200211_add_tables_for_nextjs_auth/migration.sql
```diff
@@ -0,0 +1,77 @@
...
+    "identifier" TEXT NOT NULL,
+    "token" TEXT NOT NULL,
+    "expires" TIMESTAMP(3) NOT NULL,
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updatedAt" TIMESTAMP(3) NOT NULL,
+
+    PRIMARY KEY ("id")
+);
+
+-- CreateIndex
```
feng-eric: It would be better to have a more detailed comment since the comment currently doesn't provide much additional useful information since the first few words of the sql statement are `CREATE UNIQUE INDEX`

prisma/migrations/20210529200211_add_tables_for_nextjs_auth/migration.sql
```diff
@@ -0,0 +1,77 @@
...
+CREATE UNIQUE INDEX "User.email_unique" ON "User"("email");
+
+-- CreateIndex
+CREATE UNIQUE INDEX "VerificationRequest.token_unique" ON "VerificationRequest"("token");
+
+-- CreateIndex
+CREATE UNIQUE INDEX "VerificationRequest.identifier_token_unique" ON "VerificationRequest"("identifier", "token");
+
+-- AddForeignKey
+ALTER TABLE "Account" ADD FOREIGN KEY ("userId") REFERENCES "User"("id") ON DELETE CASCADE ON UPDATE CASCADE;
```
feng-eric: I would prefer to have foreign keys be specified in the table itself so that it's easier to track. I think it would also make it more understandable if we have it in the `CREATE TABLE` statement rather than as a line by itself

prisma/migrations/migration_lock.toml
```diff
@@ -0,0 +1,3 @@
+# Please do not edit this file manually
+# It should be added in your version-control system (i.e. Git)
+provider = "postgresql"
```
feng-eric: What's this file for?

feng-eric (changes requested):
Left a few comments!

prisma/migrations/20210529200211_add_tables_for_nextjs_auth/migration.sql
```diff
@@ -0,0 +1,77 @@
+-- CreateTable
+CREATE TABLE "Account" (
```
feng-eric: Can you also explain what the difference between an `Account` entity and a `User` entity is? Is there an issue with having the fields in `User` be in Account?

prisma/migrations/20210529200211_add_tables_for_nextjs_auth/migration.sql
feng-eric: https://github.com/uwblueprint/richmond-centre-for-disability/tree/feature/initialize-tables/prisma Here is how it's done on RCD currently. Here is dancefest: https://github.com/uwblueprint/dancefest-web/tree/staging/prisma

feng-eric:
In the README, please also add the following information on how to set up DB.
```
# Deploy schema.sql to Heroku postgres
heroku pg:psql -a YOUR_APP_NAME -f prisma/schema.sql

# Regenerate Prisma schema and client
npx prisma introspect && npx prisma generate
```

pages/api/auth/[...nextauth].js
Comment (mahadzaryab1) [reply pages/api/auth/[...nextauth].js:45]: added - thanks for catching that!

pages/api/auth/[...nextauth].js
Comment (mahadzaryab1) [reply pages/api/auth/[...nextauth].js:47]: We actually don't need to do this because we have a configuration for the prisma database adapter. NextAuth will just grab the database config from the prisma adapter.

prisma/migrations/migration_lock.toml
mahadzaryab1: this is used for keeping track of if providers (ex. postgresql) were changed

prisma/migrations/20210529200211_add_tables_for_nextjs_auth/migration.sql
mahadzaryab1: These files are actually generated by prisma. I think foreign keys are created in separate statements because if they were done in link, the referenced key might not have been created yet which would result in an error

prisma/migrations/20210529200211_add_tables_for_nextjs_auth/migration.sql
mahadzaryab1: These migration files are created from the `schema.prisma` file. If we specify the limit in that file - it'll reflect in the migration file as well.

prisma/migrations/20210529200211_add_tables_for_nextjs_auth/migration.sql
mahadzaryab1: Noted - but since we don't want migrations right now this doesn't matter. Will keep this in mind for the future

prisma/migrations/20210529200211_add_tables_for_nextjs_auth/migration.sql
mahadzaryab1: Noted - but since we don't want migrations right now this doesn't matter. Will keep this in mind for the future.

prisma/migrations/20210529200211_add_tables_for_nextjs_auth/migration.sql
mahadzaryab1: This is generated by prisma - if you would like to change it I will keep that in mind for next time!

prisma/migrations/20210529200211_add_tables_for_nextjs_auth/migration.sql
mahadzaryab1: That's a good question - I was just following what the [documentation](https://next-auth.js.org/adapters/prisma) suggested. I don't think Account and Session are required though - do you think I should remove them?

feng-eric (approved):
LGTM!

PR #8, opened 2021-06-06
add prisma schema

cindy-wang328:
### Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/Set-up-DB-Schemas-13725846f2a0478da047ead88160399f)

- Put the DB schema from the ER diagram into Prisma
- [ER diagram](https://dbdiagram.io/d/60ac9538b29a09603d165d47)

### Steps to test

1.  ran npx prisma db push and npx prisma studio, created a record in the prisma UI 

image

- which fields can be optional?
- address format?
- can people from any province register (do we need the province enum)

### Checklist
Not sure, is there a linter for prisma file? I downloaded the vs code prisma extension and it said no errors
- [ x] My PR name is descriptive and in imperative tense
- [ x] I have run the linter
- [ x] I have requested a review from the PL, and/or other devs who have background knowledge on this PR or who will be building on top of this PR


prisma/schema.sql
```diff
@@ -0,0 +1,158 @@
+CREATE TYPE provinces AS ENUM('NL', 'PE', 'NS', 'NB', 'QC', 'ON', 'MB', 'SK', 'AB', 'BC', 'YT', 'NT', 'NU');
+CREATE TYPE weekdays AS ENUM ('MON', 'TUE', 'WED', 'THU', 'FRI', 'SAT', 'SUN');
+
+-- npx prisma introspect && npx prisma generate
```
feng-eric: I don't think you need this line here

prisma/schema.sql
```diff
@@ -0,0 +1,158 @@
...
+CREATE TABLE verification_requests (
+  id SERIAL PRIMARY KEY NOT NULL,
+  identifier VARCHAR(255) NOT NULL,
+  token VARCHAR(255) NOT NULL UNIQUE,
+  expires TIMESTAMPTZ NOT NULL,
+  created_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  updated_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP
+);
+
+CREATE TABLE program (
```
feng-eric: For documentation standard, please add comments on top. `-- Create program table`

prisma/schema.sql
```diff
@@ -0,0 +1,158 @@
...
+  program_id SERIAL PRIMARY KEY NOT NULL,
+  price INTEGER NOT NULL, -- cents
```
feng-eric: What does the `-- cents` comment mean?

prisma/schema.sql
feng-eric: Same for other tables too!

prisma/schema.sql
```diff
@@ -0,0 +1,158 @@
...
+  program_id INTEGER NOT NULL,
+  is_valid BOOLEAN NOT NULL,
+  FOREIGN KEY (volunteer_id) REFERENCES volunteer(volunteer_id) ON DELETE CASCADE,
+  FOREIGN KEY (parent_id) REFERENCES parent(parent_id) ON DELETE CASCADE,
+  FOREIGN KEY (student_id) REFERENCES student(student_id) ON DELETE CASCADE,
+  FOREIGN KEY (program_id) REFERENCES program(program_id) ON DELETE CASCADE,
+  PRIMARY KEY (student_id, program_id)
+);
+
+CREATE TABLE "admin" (
```
feng-eric: Why does admin have quotations around it?

cindy-wang328: its some kind of keyword, like the IDE highlighted it the same color as other keywords
victor added it he might know more

prisma/schema.sql
cindy-wang328: didnt want to deal with floating point, stackoverflow said to use the price in cents and make it an int

prisma/schema.prisma
```diff
@@ -1,12 +1,205 @@
...
-  provider          = "postgresql"
-  url               = env("DATABASE_URL")
-  shadowDatabaseUrl = env("SHADOW_DATABASE_URL")
+  provider = "postgresql"
+  url      = env("DATABASE_URL")
 }
 
-generator client {
-  provider = "prisma-client-js"
+model admin {
```
feng-eric: Please change model names to follow PascalCase, singular. This can be done by changing `admin` to `coupon_users` to `CouponUser`. Then we add `@@mapping(name: "admin")` and `@@mapping(name: "coupon_users"` at the end of each respective model.

Please see https://www.prisma.io/docs/concepts/components/prisma-schema/data-model for reference and also the `schema.prisma` file in dancefest-web repository

cindy-wang328:
ok i changed the names and added the @@map but im confused, i added all the "create program table" comments, they are in  https://github.com/uwblueprint/social-diversity-for-children/pull/8/commits/f3f1221153ac6e1f6e149bd6f2cd9846cab3752c, but they disappeared??

prisma/schema.sql
feng-eric: I see, so $50.00 is stored as 5000 right? Also when we display the price in the frontend, we'll convert it to a more standard format?

feng-eric:
> ok i changed the names and added the @@Map but im confused, i added all the "create program table" comments, they are in [f3f1221](https://github.com/uwblueprint/social-diversity-for-children/commit/f3f1221153ac6e1f6e149bd6f2cd9846cab3752c), but they disappeared??

That's really strange. @mahadzaryab1, have you had this issue before?

mahadzaryab1:
> > ok i changed the names and added the @@Map but im confused, i added all the "create program table" comments, they are in [f3f1221](https://github.com/uwblueprint/social-diversity-for-children/commit/f3f1221153ac6e1f6e149bd6f2cd9846cab3752c), but they disappeared??
> 
> That's really strange. @mahadzaryab1, have you had this issue before?



> > ok i changed the names and added the @@Map but im confused, i added all the "create program table" comments, they are in [f3f1221](https://github.com/uwblueprint/social-diversity-for-children/commit/f3f1221153ac6e1f6e149bd6f2cd9846cab3752c), but they disappeared??
> 
> That's really strange. @mahadzaryab1, have you had this issue before?

I don't think so - the prisma commands shouldn't affect the `prisma.sql` file because that's written by us.

prisma/schema.sql
mahadzaryab1: can we make this comment more descriptive?

prisma/schema.sql
mahadzaryab1: should we add a `created_at` and `updated_at` field to this table as well as all other table for tracking purposes?

prisma/schema.sql
```diff
@@ -0,0 +1,170 @@
...
+  last_name VARCHAR(255) NOT NULL,
+  phone_number VARCHAR(50) NOT NULL,
+  email VARCHAR(255) UNIQUE NOT NULL,
+  is_low_income BOOLEAN DEFAULT false,
+  address_line1 VARCHAR(255) NOT NULL,
+  address_line2 VARCHAR(255),
+  postal_code VARCHAR(10) NOT NULL,
+  city_name VARCHAR(50) NOT NULL,
+  province provinces NOT NULL,
+  preferred_language VARCHAR(20) NOT NULL
```
mahadzaryab1: can this be turned into an enum? we have a handful of languages we're planning to support, right?

prisma/schema.sql
```diff
@@ -0,0 +1,170 @@
...
+);
+
+-- create volunteer table
+CREATE TABLE volunteer (
+  volunteer_id SERIAL PRIMARY KEY NOT NULL,
+  first_name VARCHAR(255)  NOT NULL,
```
mahadzaryab1: I think we can set names as well as other fields of unknown sizes as `TEXT`. Thoughts? @feng-eric

mahadzaryab1 (commented):
Added a few comments/questions

prisma/schema.sql
cindy-wang328: ok https://stackoverflow.com/questions/4848964/difference-between-text-and-varchar-character-varying says varchar and text are stored the same way and same performance

prisma/schema.sql
feng-eric: Yea we should. I thought Prisma would auto generate those ones actually but perhaps I remembered incorrectly

prisma/schema.sql
feng-eric: Yeah, I think either way works. I'm going to assume that we won't ever be storing a name more than 255 characters.

cindy-wang328:
why are there so many files changed :(

prisma/schema.sql
cindy-wang328: i made all varchar(255)s text

prisma/schema.sql
```diff
@@ -0,0 +1,194 @@
...
+  parent_id INTEGER NOT NULL,
+  FOREIGN KEY (parent_id) REFERENCES parent(parent_id) ON DELETE CASCADE,
+  FOREIGN KEY (program_id) REFERENCES program(program_id) ON DELETE CASCADE,
+  PRIMARY KEY (parent_id, program_id),
+  created_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  updated_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP
+);
+-- create student table
+CREATE TABLE student(
+  student_id SERIAL PRIMARY KEY NOT NULL,
```
feng-eric: I think for tables with only one id, you can just call it `id`. I noticed some tables have this but some tables with only 1 id also have the table name in the field name too. Let's try to keep it consistent

prisma/schema.sql
```diff
@@ -0,0 +1,194 @@
...
+  'MB',
+  'SK',
+  'AB',
+  'BC',
+  'YT',
+  'NT',
+  'NU'
+);
+CREATE TYPE weekdays AS ENUM ('MON', 'TUE', 'WED', 'THU', 'FRI', 'SAT', 'SUN');
+-- https://stackoverflow.com/questions/3191664/list-of-all-locales-and-their-short-codes
```
feng-eric, line 17: Good comment!

prisma/schema.sql
cindy-wang328: but then if we have to join a lot of tables that all have ids there will be ambiguous reference like you dont know whose id it is
and then we will have to do more renames or "tablename.id"

feng-eric: In Prisma, you don't have to actually write "joins". For example, in the screenshot below, in order to access the data of parent, it's accessed similarly like any other field. There is never a need to actually write raw sql when querying. If we change to just `id`, it will just change to `@relation(fields: [parent_id], references: [id]`. There isn't a need to rename or do "tablename.id".

```
model ParentOfStudent {
  parent_id  Int
  student_id Int
  created_at DateTime @default(now()) @db.Timestamptz(6)
  updated_at DateTime @default(now()) @db.Timestamptz(6)
  parent     Parent   @relation(fields: [parent_id], references: [parent_id])
  student    Student  @relation(fields: [student_id], references: [student_id])
  @@id([student_id, parent_id])
  @@map("parent_of_students")
}
```

feng-eric: For example, this is a query where we join another table (from dancefest). By just having the `include`, we're able to access relational data. 

```
return await prisma.event.findUnique({    
    where: {      
        id: parseInt(id),    
    },    
    include: {      
        performances: true,    
    },  
});
```

prisma/schema.sql
```diff
@@ -0,0 +1,194 @@
...
+  student_id INTEGER NOT NULL,
+  FOREIGN KEY (parent_id) REFERENCES parents(id) ON DELETE CASCADE,
+  FOREIGN KEY (student_id) REFERENCES students(id) ON DELETE CASCADE,
+  PRIMARY KEY (student_id, parent_id),
+  created_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  updated_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP
+);
+-- create table for registration with parents and students
+CREATE TABLE parent_regs (
+  volunteer_id INTEGER NOT NULL,
```
feng-eric: What's the reason for why `volunteer_id` is here?

feng-eric (approved):
Just had one quick comment I put but other than that, it looks good! Thanks a lot for all the hardwork @cindy-wang328 @sohak22 @victoranyun!!  🙌

mahadzaryab1 (approved):
LGTM - great work!!

PR #11, opened 2021-06-13
Set up s3 file upload

jason17h:
### Notion ticket link

[Set up S3 upload](https://www.notion.so/uwblueprintexecs/086c1b564dbf478faa649a8880e34b81?v=9d3f8002a0d1477ba395616dfc082af7&p=73e58b47c00c457a96e5592ffd71ac24)

Implementing function to upload a file to S3.

### Steps to test

Called the function using a sample file as input and confirmed on the AWS console that the file was properly uploaded.

-

### Checklist

-   [x] I have run the linter


utils/uploadToS3.ts
```diff
@@ -0,0 +1,29 @@
+import aws from "aws-sdk";
+import fs from "fs";
+import path from "path";
+
```
feng-eric: Add comment to document function

utils/uploadToS3.ts
```diff
@@ -0,0 +1,29 @@
...
+const uploadToS3 = (filePath: string): void => {
+    aws.config.update({
+        accessKeyId: process.env.AWS_ACCESS_KEY,
+        secretAccessKey: process.env.AWS_SECRET_KEY,
+        region: process.env.AWS_REGION,
+        signatureVersion: "v4",
+    });
+    const s3 = new aws.S3();
```
feng-eric: I think `new aws.S3()` should be exported to this file so we don't create new s3 instances

utils/uploadToS3.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    const fileStream = fs.createReadStream(filePath);
+    const uploadParams = {
+        Bucket: process.env.AWS_BUCKET_NAME,
+        Key: path.basename(filePath),
+        Body: fileStream,
+    };
+    s3.upload(uploadParams, function (err, data) {
+        if (err) {
+            console.log("ERROR: ", err);
```
mahadzaryab1: return error rather than logging

utils/uploadToS3.ts
```diff
@@ -0,0 +1,29 @@
...
+        } else {
+            console.log("SUCCESS: ", data.Location);
```
mahadzaryab1: return success object rather than logging

utils/uploadToS3.ts
```diff
@@ -0,0 +1,29 @@
...
+        }
+    });
+};
+
+export default uploadToS3;
```
mahadzaryab1: can you add an example of how this function can/should be called

mahadzaryab1 (changes requested):
Added a few comments. Another thing to consider is that your file is called `uploadToS3.ts` but I think it should maybe just be `s3.ts` or `s3Utility.ts` so that any s3 utility functions can be added rather than just the upload.

utils/uploadToS3.ts
jason17h: see https://github.com/uwblueprint/social-diversity-for-children/pull/11/commits/5652c943c9e3b0fb96cab393fd2dfe0a93275e68

utils/uploadToS3.ts
jason17h: I assume you meant exported _from_ this file? If so, see https://github.com/uwblueprint/social-diversity-for-children/pull/11/commits/d683e84f081d2625d8dce69dbf3c462f5be123c8
Otherwise, let me know what you meant

feng-eric: After #12 is merged in, perhaps you could set up & export  `s3` object in `aws/index.js`.

utils/aws/s3.ts
```diff
@@ -0,0 +1,41 @@
...
+    * returns: upload response if successful, error object otherwise
+
+Sample usage:
+    myFile = "~/documents/my_file.txt"
+    res = uploadToS3(myFile)
+
+*/
+
+const uploadToS3 = (filePath: string): void => {
+    aws.config.update({
```
feng-eric: I don't think this code needs to be in this function here. It can be set outside it since there isn't a need to update the config each time we upload to s3. I think you can have this in `aws.index.js`. See my other comment!

utils/uploadToS3.ts
feng-eric: Actually, I think it might be okay to merge yours in first. Perhaps you can create an `index.js` file where you set up everything from aws. See `aws/index.js` in https://github.com/uwblueprint/social-diversity-for-children/pull/12/files and https://github.com/uwblueprint/dancefest-web/blob/staging/aws/index.js

utils/aws/s3.ts
```diff
@@ -0,0 +1,41 @@
...
+    };
+    S3.upload(uploadParams, function (err, data) {
+        if (err) {
+            return err;
+        }
+        return data;
+    });
+};
+
+export { S3, uploadToS3 };
```
mahadzaryab1: @feng-eric - why does `S3` need to be exported? is it going to be used outside this file?

utils/aws/s3.ts
```diff
@@ -0,0 +1,41 @@
...
+    * filePath: the path of the file to upload
+    * returns: upload response if successful, error object otherwise
+
+Sample usage:
+    myFile = "~/documents/my_file.txt"
+    res = uploadToS3(myFile)
+
+*/
+
+const uploadToS3 = (filePath: string): void => {
```
mahadzaryab1: the function shouldn't be returning void - it should maybe return some type of response object that indicates if the upload was successful, error messages, and potentially the data object

utils/aws/s3.ts
```diff
@@ -0,0 +1,41 @@
...
+
+Uploads a file to the AWS S3 bucket corresponding to AWS_BUCKET_NAME in .env.
+    * filePath: the path of the file to upload
+    * returns: upload response if successful, error object otherwise
+
+Sample usage:
+    myFile = "~/documents/my_file.txt"
+    res = uploadToS3(myFile)
+
+*/
```
mahadzaryab1: Use jsdoc syntax for function documentation - see [here](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html)

utils/aws/s3.ts
```diff
@@ -0,0 +1,41 @@
...
+
+/* 
+
+Uploads a file to the AWS S3 bucket corresponding to AWS_BUCKET_NAME in .env.
+    * filePath: the path of the file to upload
+    * returns: upload response if successful, error object otherwise
+
+Sample usage:
+    myFile = "~/documents/my_file.txt"
+    res = uploadToS3(myFile)
```
mahadzaryab1: Have you tried this to make sure it works? it might be helpful to test this if you haven't already on some test endpoint.

mahadzaryab1 (changes requested):
Added a few more comments!

utils/aws/s3.ts
feng-eric: My initial comment was saying we can initialize and export s3 object from `aws/index.js`. We can also just initialize in here outside of the function instead too. Either one works.

utils/aws/s3.ts
feng-eric: I think it would be most useful to return the response from the s3 upload.

utils/aws/s3.ts
feng-eric: I didn't mean export it out of this file

utils/aws/s3.ts
brandonwong368: Yeah, we called the function in an endpoint and got a sample text file to upload

utils/uploadToS3.ts
jason17h: Since the success object is in the callback function, can it be returned from the calling function?

utils/aws/s3.ts
jason17h: Is this what you had in mind? https://github.com/uwblueprint/social-diversity-for-children/pull/11/commits/4156ac1af5483819aca96900b2314dc72f6a486e

utils/uploadToS3.ts
jason17h: See https://github.com/uwblueprint/social-diversity-for-children/pull/11/commits/4156ac1af5483819aca96900b2314dc72f6a486e

mahadzaryab1 (changes requested):
Added a few more comments

yarn.lock
```diff
@@ -4631,22 +4631,14 @@ buffer@4.9.2, buffer@^4.3.0:
...
+buffer@5.6.0, buffer@^5.5.0:
   version "5.6.0"
-  resolved "https://registry.yarnpkg.com/buffer/-/buffer-5.6.0.tgz#a31749dc7d81d84db08abf937b6b8c4033f62786"
+  resolved "https://registry.npmjs.org/buffer/-/buffer-5.6.0.tgz"
   integrity sha512-/gDYp/UtU0eA1ys8bOs9J6a+E/KWIY+DZ+Q2WESNUA0jFRsJOc0SNUO6xJ5SGA1xueg3NL65W6s+NY5l9cunuw==
   dependencies:
     base64-js "^1.0.2"
     ieee754 "^1.1.4"
 
-buffer@^5.5.0:
```
brandonwong368: I was trying to resolve the conflicts with staging and I think I removed a dependency by accident @mahadzaryab1 @feng-eric, do you know which aws-sdk, buffer, and ieee versions we are supposed to keep

utils/aws/s3.ts
```diff
@@ -0,0 +1,27 @@
...
+        Bucket: process.env.AWS_BUCKET_NAME,
+        Key: path.basename(filePath),
+        Body: fileStream,
+    };
+
+    const res = s3.upload(uploadParams, function (err, data) {
+        console.log(err, data);
+    });
+
+    return res;
```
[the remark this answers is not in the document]
jason17h: Updated it so that it now returns a promise which the client can work with: https://github.com/uwblueprint/social-diversity-for-children/pull/11/commits/8798d6abefc54c21c27187d32cbade845b641d21

package.json
```diff
@@ -50,7 +50,6 @@
         "@typescript-eslint/parser": "^4.9.1",
         "babel-loader": "^8.2.2",
         "css-loader": "^5.0.1",
-        "dotenv": "^8.2.0",
```
feng-eric, line 47: How come this was removed?

feng-eric: Already talked to @brandonwong368 about this but let's also remove `package-lock.json`

aws/index.ts
```diff
@@ -1,11 +1,12 @@
 import aws from "aws-sdk";
 import nodemailer from "nodemailer";
 
-// write to aws config
+// Update AWS Config
 aws.config.update({
     accessKeyId: process.env.AWS_ACCESS_KEY_ID,
     secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY,
     region: process.env.AWS_REGION,
+    signatureVersion: "v4",
```
feng-eric, line 9: Is this required?

feng-eric (approved):
LGTM! Before merging in, let's get @mahadzaryab1 to take a look too and also if you guys could address the comment I put.

aws/index.ts
brandonwong368: It's not required, I tested the function with an API endpoint without `signatureVersion: "v4",` and the upload worked fine. 

I took this from the AWS documentation: The Signature Version 2 or Signature Version 4 signing process is used to authenticate your Amazon S3 API requests. Signing requests enables Amazon S3 to identify who is sending the request and protects your requests from bad actors. https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingAWSSDK.html

It seems like removing it could have a security issue later on? Not sure, I won't remove it until @mahadzaryab1  takes a look

mahadzaryab1: I think this is good to keep due to the reasons that @brandonwong368 suggested.

utils/aws/s3.ts
```diff
@@ -0,0 +1,29 @@
...
+import fs from "fs";
+import path from "path";
+
+/**
+ * Uploads a file to the AWS S3 bucket corresponding to env.AWS_BUCKET_NAME
+ * @param {string} bucketName - Name of the bucket to upload the file to
+ * @param {string} filePath - Path of the file to upload
+ * @returns {promise} - S3 upload response object on resolve, error object on reject
+ */
+const uploadToS3 = (bucketName: string, filePath: string): Promise<unknown> => {
```
mahadzaryab1: This function actually isn't returning a promise anymore because you're resolving the promise within the function itself. It seems to either return `data` or `err` which or of two different types and you generally want to avoid doing that in Typescript. You have two options here: 
- Instead of resolving the promise within the function itself, you can return it from the function and let the caller resolve it. 
- Make a custom s3 response object similar to [this](https://github.com/uwblueprint/social-diversity-for-children/blob/staging/src/models/response.ts#L6-L9) and then the caller can check to see if the error field is null or not. If it's not null there's an error and if it is then there's no error and the caller can check the data field.  

The above two suggestions can be done in separate PR if you like but for now it's probably good to change the return type from `Promise<unknown>` to just `unknown`. cc @feng-eric for thoughts

utils/aws/s3.ts
```diff
@@ -0,0 +1,29 @@
+import { s3 } from "aws/index";
+import fs from "fs";
+import path from "path";
+
+/**
+ * Uploads a file to the AWS S3 bucket corresponding to env.AWS_BUCKET_NAME
```
mahadzaryab1: nit: change comment because the bucket doesn't correspond to `AWS_BUCKET_NAME` anymore from the perspective of this file

mahadzaryab1 (changes requested):
Overall, looks great - nice work!! I just highlighted one thing that needs to be fixed and we should be good to go after!

utils/aws/s3.ts
feng-eric: I think the second approach with using a custom s3 response object seems good. We can do it in a different PR.

utils/aws/s3.ts
mahadzaryab1: did you do this ?

mahadzaryab1 (approved):
LGTM - just fix the comment on the function! Great work!!

PR #19, opened 2021-06-26
Update AWS Service To Add More Calls For Interacting With S3

mahadzaryab1:
### Notion ticket link

[S3 Functions](https://www.notion.so/uwblueprintexecs/S3-Functions-7ccea6903d874b38b893dbf88ef1cc37)

- Refactored the existing `uploadFileToS3` function to return a promise 
- Added a function `getFileFromS3` to retrieve a file from S3
- Added a function `deleteFileFromS3` to delete a file from 3

#### Retrieving an Object
```typescript
    getFileFromS3("some-bucket-name", "some-file-key")
        .then((data) => {
            console.log(data.Body.toString());
        })
        .catch((err) => {
            console.log(err);
        });
```
#### Uploading an Object
```typescript
    uploadFileToS3("some-bucket-name", "some-file-path")
        .then((data) => {
            console.log(data);
        })
        .catch((err) => {
            console.log(err);
        });
```

#### Deleting an Object
```typescript
    deleteFileFromS3("some-bucket-name", "some-file-key")
        .then((data) => {
            console.log(data);
        })
        .catch((err) => {
            console.log(err);
        });
```

### Checklist

-   [x] I have run the linter

### DIFF
--- a/services/aws/s3.ts
+++ b/services/aws/s3.ts
@@ -1,16 +1,47 @@
 import { s3 } from "services/aws/index";
 import fs from "fs";
 import path from "path";
-import { ManagedUpload } from "aws-sdk/clients/s3";
+import S3, { ManagedUpload } from "aws-sdk/clients/s3";
+import { AWSError } from "aws-sdk";
+
+/**
+ * Retrieves a file from a S3 Bucket
+ * @param {string} bucketName Name of the bucket to retrieve the file from
+ * @param {string} fileKey The key of the file in the bucket
+ * @returns {S3.GetObjectOutput} A promise which can be resolved to obtain the file from S3 and catch
+ * any errors in making the get call
+ */
+const getFileFromS3 = (
+    bucketName: string,
+    fileKey: string,
+): Promise<S3.GetObjectOutput> => {
+    return new Promise<S3.GetObjectOutput>((resolve, reject) => {
+        const retrieveParams = {
+            Bucket: bucketName,
+            Key: fileKey,
+        };
+
+        s3.getObject(
+            retrieveParams,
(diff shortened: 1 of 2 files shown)


services/aws/s3.ts
```diff
@@ -1,29 +1,105 @@
 import { s3 } from "services/aws/index";
 import fs from "fs";
 import path from "path";
+import S3, { ManagedUpload } from "aws-sdk/clients/s3";
```
feng-eric: Do you think having `s3` and `S3` is a bit confusing?

services/aws/s3.ts
```diff
@@ -1,29 +1,105 @@
...
+import { AWSError } from "aws-sdk";
```
feng-eric: I would import this from `serviecs/aws/index`.I think it would be cleaner if we import everything from `aws-sdk` in one file. Thoughts?

feng-eric: Especially if there might be more AWS services that might use this

feng-eric (approved):
Looks really great! Thanks for the hard work. Had a few comments/thoughts about importing from `aws`. They aren't too big of a deal.

Were you able to test this manually?

services/aws/s3.ts
mahadzaryab1: good idea - done!

mahadzaryab1:
> Looks really great! Thanks for the hard work. Had a few comments/thoughts about importing from `aws`. They aren't too big of a deal.
> 
> Were you able to test this manually?

Yeah! I tested it using the usages I mentioned in the PR description.

feng-eric (approved):
LGTM!

PR #23, opened 2021-06-28
add program sections table, attempt 2

cindy-wang328:

Designers said each program should have different sections, with different days/times and age groups
image

### Steps to test

1. reset database on heroku website
2. heroku pg:psql -a sdc-dev-2 -f prisma/schema.sql
3. npx prisma introspect, npx prisma generate
4. change variable names in prisma file and add @@map 

- is there any info in the figma that is still missing
- maybe forgot some NOT NULL

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -23,7 +23,6 @@ model Parent {
   first_name         String
   last_name          String
   phone_number       String            @db.VarChar(50)
-  email              String            @unique
   is_low_income      Boolean?          @default(false)
   address_line1      String
   address_line2      String?
@@ -33,6 +32,7 @@ model Parent {
   preferred_language locales
   created_at         DateTime          @default(now()) @db.Timestamptz(6)
   updated_at         DateTime          @default(now()) @db.Timestamptz(6)
+  users              User              @relation("parentsTousers", fields: [id], references: [id])
   coupon_users       CouponUsers[]     @relation("coupon_usersToparents")
   parent_of_students ParentOfStudent[] @relation("parent_of_studentsToparents")
   parent_regs        ParentReg[]       @relation("parent_regsToparents")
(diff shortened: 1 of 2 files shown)


prisma/schema.prisma
```diff
@@ -23,6 +23,7 @@ model Parent {
   first_name         String
   last_name          String
   phone_number       String            @db.VarChar(50)
+  email              String            @unique
```
feng-eric: This should be removed from the parent table because we have it in the users table

prisma/schema.sql
```diff
@@ -13,13 +13,11 @@ CREATE TYPE provinces AS ENUM(
...
   'NU'
 );
-
 CREATE TYPE weekdays AS ENUM ('MON', 'TUE', 'WED', 'THU', 'FRI', 'SAT', 'SUN');
-
 -- https://stackoverflow.com/questions/3191664/list-of-all-locales-and-their-short-codes
 -- chinese, english, japanese, korean
 CREATE TYPE locales AS ENUM ('zh', 'en', 'ja', 'ko');
-
+CREATE TYPE online_formats AS ENUM ('online', 'in-person', 'blended');
```
feng-eric: I think this should be called program_formats

prisma/schema.sql
```diff
@@ -65,23 +83,23 @@ CREATE TABLE parents (
   first_name TEXT NOT NULL,
   last_name TEXT NOT NULL,
   phone_number VARCHAR(50) NOT NULL,
+  email TEXT UNIQUE NOT NULL,
```
feng-eric: Let's remove this as it's in user table

prisma/schema.sql
```diff
@@ -65,23 +83,23 @@ CREATE TABLE parents (
...
   is_low_income BOOLEAN DEFAULT false,
   address_line1 TEXT NOT NULL,
   address_line2 TEXT,
   postal_code VARCHAR(10) NOT NULL,
   city_name VARCHAR(50) NOT NULL,
   province provinces NOT NULL,
   preferred_language locales NOT NULL,
-  FOREIGN KEY(id) REFERENCES users(id),
```
feng-eric, line 75: How come this was removed?

prisma/schema.sql
```diff
@@ -65,23 +83,23 @@ CREATE TABLE parents (
...
   created_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
   updated_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP
 );
-
 -- create volunteer table
 CREATE TABLE volunteers (
   id SERIAL PRIMARY KEY NOT NULL,
   first_name TEXT NOT NULL,
   last_name TEXT NOT NULL,
+  email TEXT UNIQUE,
```
feng-eric: Same as above

prisma/schema.sql
```diff
@@ -91,23 +109,20 @@ CREATE TABLE volunteers (
   city_name TEXT,
   province provinces,
   preferred_language locales,
-  FOREIGN KEY(id) REFERENCES users(id),
```
feng-eric, line 94: Same as above

feng-eric (changes requested):
Had some comments!

prisma/schema.sql
```diff
@@ -40,21 +37,42 @@ CREATE TABLE verification_requests (
...
+
+  start_date TIMESTAMPTZ NOT NULL, -- earliest start date of all sections
+  end_date TIMESTAMPTZ NOT NULL, -- latest end date of all sections
+
+  is_archived BOOLEAN DEFAULT false NOT NULL,
+  created_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  updated_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP
+);
+
+CREATE TABLE sections (
```
feng-eric: We should have `ON DELETE CASCADE` so that if a program gets deleted, the associating records here also get deleted.

feng-eric:
@cindy-wang328 The designers came to the conclusion that they want to call it `classes` instead of `sections`. Could we make that change in this PR?

cindy-wang328:
> @cindy-wang328 The designers came to the conclusion that they want to call it `classes` instead of `sections`. Could we make that change in this PR?

would there be any problems with confusing the OOP "class" with naming these models/tables "class"? like at work in the codebase theres some places that use "clazz" bc they cant use class as a variable name

prisma/schema.prisma
```diff
@@ -260,3 +290,9 @@ enum weekdays {
   SAT
   SUN
 }
+
+enum program_formats {
+  online
+  in_person @map("in-person")
```
feng-eric, line 296: there's actually a ticket to change all prisma fields to camelCase lol

feng-eric: you can ignore, we can do it in a separate PR

feng-eric (approved):
LGTM!

feng-eric:
This might break some stuff with the GET endpoint

PR #25, opened 2021-06-29
Implement delete program endpoint

brandonwong368:
### Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/086c1b564dbf478faa649a8880e34b81?v=9d3f8002a0d1477ba395616dfc082af7&p=eb4f61da390f40b98d2321e90a5b3126)

- created deleteProgram function to delete a record of corresponding id
- added the handling for DELETE request in [id].ts

### Steps to test

1.

-

### Checklist

-   [x] I have run the linter


brandonwong368:
Should I add a function for deleting all records in the database? ie. `prisma.program.deleteMany()`

feng-eric:
Not sure what's different from the other PR so I'll review the other program PRs first so I'm not doing duplicate reviews

feng-eric:
> Should I add a function for deleting all records in the database? ie. `prisma.program.deleteMany()`

I don't think you need it as I don't see a use case for deleting multiple.

pages/api/program/index.ts
```diff
@@ -0,0 +1,65 @@
...
+                }
+                ResponseUtil.returnOK(res, newProgram);
+                break;
+            }
+        }
+        case "PUT": {
+            // TODO:
+            break;
+        }
+        case "DELETE": {
```
feng-eric: You can probably get rid of this since I don't see a need for a delete many program endpoint

pages/api/program/index.ts
```diff
@@ -0,0 +1,65 @@
...
+            // TODO:
+            break;
+        }
+        default: {
+            const allowedHeaders: string[] = ["GET", "POST", "PUT", "DELETE"];
```
feng-eric:
```suggestion
            const allowedHeaders: string[] = ["GET", "POST", "PUT"];
```

pages/api/program/index.ts
```diff
@@ -0,0 +1,65 @@
...
+            // TODO: add JSON response for method not allowed
```
feng-eric:
```suggestion
```

services/database/program.ts
```diff
@@ -0,0 +1,67 @@
...
+        },
+    });
+    return program;
+}
+
+/**
+ * deleteProgram takes the id parameter and deletes the program associated with the programId
+ * @param {string} id - programId
+ */
+
```
feng-eric:
```suggestion
```

utils/validation/program.ts
```diff
@@ -0,0 +1,29 @@
+import { createProgramInput } from "models/Program";
+
+/**
+ * validateProgram takes programData of type Program and validates the data
+ * @param newProgramData - data corresponding to new program
+ * @returns true: validated, false: not validated
+ */
+export function validateCreateProgram(
```
feng-eric, line 8: I think I said this in another PR but it would be good to return a specific error regarding what was invalid about the input. This way, it'll provide a much more descriptive error message to the frontend.

feng-eric (changes requested):
left some comments. Overall, it looks super clean btw! Great work :)

models/Program.ts
```diff
@@ -0,0 +1,23 @@
...
+    start_date: string;
+    end_date: string;
+    weekday: weekdays;
+    start_time_minutes: number;
+    duration_minutes: number;
+    space_total: number;
+    space_available: number;
+    volunteer_space_total: number;
+    volunteer_space_available: number;
+    is_archived?: boolean;
```
mahadzaryab1: nit: ts prefers camelCase instead of snake_case

models/Program.ts
```diff
@@ -0,0 +1,23 @@
...
+    MON = "MON",
+    TUE = "TUE",
+    WED = "WED",
+    THU = "THU",
+    FRI = "FRI",
+    SAT = "SAT",
+    SUN = "SUN",
+}
+
+export type createProgramInput = {
```
mahadzaryab1, line 18: also - add type documentation because it will be likely be used by both fe/be developers

models/Program.ts
```diff
@@ -0,0 +1,23 @@
+export enum weekdays {
```
mahadzaryab1, line 4: add documentation

pages/api/program/[id].ts
```diff
@@ -0,0 +1,54 @@
...
+ */
+export default async function handle(
+    req: NextApiRequest,
+    res: NextApiResponse,
+): Promise<void> {
+    if (req.method == "GET") {
+        // Obtain program id
+        const { id } = req.query;
+        // Obtain program with provided programId
+        const program = await getProgram(id as string);
```
Comment (mahadzaryab1) [inline pages/api/program/[id].ts:20]: add type for `program`

services/database/program.ts
```diff
@@ -0,0 +1,67 @@
...
+async function getProgram(id: string): Promise<Program> {
+    const program = await prisma.program.findUnique({
+        where: {
+            id: parseInt(id),
+        },
+    });
+    return program;
+}
+
+async function getPrograms(): Promise<Program[]> {
```
mahadzaryab1: add function documentation

pages/api/program/[id].ts
```diff
@@ -0,0 +1,54 @@
...
+        // Delete program with provided programId
+        const program = await deleteProgram(id as string);
+
+        if (!program) {
+            ResponseUtil.returnNotFound(
+                res,
+                `Program with id ${id} not found.`,
+            );
+            return;
+        }
```
Comment (mahadzaryab1) [inline pages/api/program/[id].ts]: I wonder if we should return a NOT FOUND for a DELETE request that doesn't exist because that means it's already "deleted". thoughts? @feng-eric

mahadzaryab1 (commented):
added a few comments

pages/api/program/[id].ts
Comment (feng-eric) [reply pages/api/program/[id].ts]: Can you explain what you mean by return a NOT FOUND?

Also, I think there's also the case where we delete something that wasn't there in the first place (doesn't have to be already "deleted")

pages/api/program/[id].ts
Comment (brandonwong368) [reply pages/api/program/[id].ts:20]: this was something in the GET endpoint that was just merged in, can we add an issue and fix it separately?

brandonwong368:
I thought it would be better to merge PR #26 first to deal with the changes to the prisma schema, but I read the comments in #26 and it says these endpoints should be merged first. What should we do?

models/Program.ts
```diff
@@ -0,0 +1,30 @@
+/**
+ * custom type enumerating days of the week for the createProgramInput custom type
+ */
```
jason17h, line 3: for enums like `provinces` and `roles`, I imported them directly from `@prisma/client` so that what's stored in the database is always consistent with what we're using in our codebase. What are your thoughts on that approach v.s. defining it here, since we should probably be consistent? @feng-eric @mahadzaryab1

pages/api/program/[id].ts
```diff
@@ -15,10 +15,25 @@ export default async function handle(
     if (req.method == "GET") {
         // Obtain program id
         const { id } = req.query;
-
-        // obtain program with provided programId
+        // Obtain program with provided programId
+        //TODO: add type for program
```
Comment (jason17h) [inline pages/api/program/[id].ts:19]: looks like this TODO has been addressed?

models/Program.ts
jason17h: seems like the TypeScript convention is to capitalize types: https://www.typescriptlang.org/docs/handbook/2/objects.html

models/Program.ts
```diff
@@ -0,0 +1,30 @@
...
+export enum weekdays {
+    MON = "MON",
+    TUE = "TUE",
+    WED = "WED",
+    THU = "THU",
+    FRI = "FRI",
+    SAT = "SAT",
+    SUN = "SUN",
```
mahadzaryab1, line 11: do we need strings here?

services/database/program.ts
```diff
@@ -23,4 +24,44 @@ async function getPrograms(): Promise<Program[]> {
...
+        },
+    });
+    return program;
+}
+
+/**
+ * deleteProgram takes the id parameter and deletes the program associated with the programId
+ * @param {string} id - programId
+ */
+async function deleteProgram(id: string): Promise<Program> {
```
mahadzaryab1, line 58: @feng-eric - same question about hard vs. soft deletes here

mahadzaryab1 (commented):
Added a question for @feng-eric. Also, I think this PR does more than just the delete endpoint so consider changing the PR name.

services/database/program.ts
feng-eric: Let's do hard delete for now and get the PR in. If we need soft delete later, we can make that change. Let's discuss with the entire team if there's a necessary use case for soft delete.

models/Program.ts
feng-eric: > Added a question for @feng-eric. Also, I think this PR does more than just the delete endpoint so consider changing the PR name.

It's because Brandon branched off of another local branch so there are changes that he didn't make. It's a bit confusing yeah

feng-eric (commented):
Seems to have some unaddressed comments + failing build currently

models/Program.ts
feng-eric: Yes, I think that's definitely better.

PR #27, opened 2021-07-01
Add GET, POST, DELETE Class Endpoints

feng-eric:
### Notion ticket link

[Create CRUD endpoints for Class](https://www.notion.so/uwblueprintexecs/086c1b564dbf478faa649a8880e34b81?v=9d3f8002a0d1477ba395616dfc082af7&p=f5791788919c4dee9753bb3117b1106b)

- Added get class by id, get classes, delete class, and create class endpoints.
- PUT endpoint still needs to be done.
- We require more validation for createClassInput but that can be done at a later time.

### Steps to test

1.

-

### Checklist

-   [ ] I have run the linter

### DIFF
--- a/models/Class.ts
+++ b/models/Class.ts
@@ -1,0 +1,28 @@
+/**
+ * Enum for the weekday field of class
+ */
+export enum weekdays {
+    MON = "MON",
+    TUE = "TUE",
+    WED = "WED",
+    THU = "THU",
+    FRI = "FRI",
+    SAT = "SAT",
+    SUN = "SUN",
+}
+
+/**
+ * Request Body Input for POST /class
+ */
+export type createClassInput = {
+    name: string;
+    ageGroup: string;
+    programId: number;
+    spaceTotal: number;
+    volunteerSpaceTotal: number;
+    startDate: string;
+    endDate: string;
+    weekday: weekdays;
+    startTimeMinutes: number;
+    durationMinutes: number;
+};
--- a/models/Class.ts
+++ b/models/Class.ts
@@ -15,8 +15,8 @@ export enum Weekday {
  * Request Body Input for POST /class
  */
 export type CreateClassInput = {
-    name: string;
-    ageGroup: string;
+    name?: string;
+    ageGroup?: string;
     programId: number;
     spaceTotal: number;
     volunteerSpaceTotal: number;
--- a/models/Class.ts
+++ b/models/Class.ts
@@ -1,7 +1,7 @@
 /**
  * Enum for the weekday field of class
  */
-export enum weekdays {
+export enum Weekday {
     MON = "MON",
     TUE = "TUE",
     WED = "WED",
@@ -14,7 +14,7 @@ export enum weekdays {
 /**
  * Request Body Input for POST /class
  */
-export type createClassInput = {
+export type CreateClassInput = {
     name: string;
     ageGroup: string;
     programId: number;
@@ -22,7 +22,7 @@ export type createClassInput = {
     volunteerSpaceTotal: number;
     startDate: string;
     endDate: string;
-    weekday: weekdays;
+    weekday: Weekday;
     startTimeMinutes: number;
     durationMinutes: number;
 };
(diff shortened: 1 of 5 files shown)


models/Class.ts
```diff
@@ -0,0 +1,28 @@
...
+    FRI = "FRI",
+    SAT = "SAT",
+    SUN = "SUN",
+}
+
+/**
+ * Request Body Input for POST /class
+ */
+export type createClassInput = {
+    name: string;
```
feng-eric: Add `?` to the fields that are optional

pages/api/class/[id].ts
```diff
@@ -0,0 +1,45 @@
...
+        const deletedClass = await deleteClass(id as string);
+
+        if (!deleteClass) {
+            ResponseUtil.returnNotFound(res, `Class with id ${id} not found.`);
+            return;
+        }
+        ResponseUtil.returnOK(res, deletedClass);
+        return;
+    } else {
+        const allowedHeaders: string[] = ["GET"];
```
Comment (jason17h) [inline pages/api/class/[id].ts:37]: does `"DELETE"` need to be added here?

services/database/class.ts
```diff
@@ -0,0 +1,66 @@
+import prisma from "@database";
+import { Class } from "@prisma/client";
+import { createClassInput } from "models/Class";
```
jason17h: Is it generally better convention to have these input types be capitalized or lowercase at the start? I went with uppercase since that seems to be the case for all the types defined in `@prisma/client` and we should be consistent

services/database/class.ts
```diff
@@ -0,0 +1,66 @@
...
+ * @returns Promise<Class> - Promise with the deleted class
+ */
+async function deleteClass(id: string): Promise<Class> {
+    const deletedClass = await prisma.class.delete({
+        where: {
+            id: parseInt(id),
+        },
+    });
+    return deletedClass;
+}
```
mahadzaryab1, line 64: consideration: should we be doing a hard delete or a soft delete?

mahadzaryab1 (approved):
One question/comment - otherwise LGTM

pages/api/class/[id].ts
Comment (feng-eric) [reply pages/api/class/[id].ts:37]: good catch!

services/database/class.ts
feng-eric: I'm not really sure what the convention is, perhaps @mahadzaryab1 could give more thoughts

services/database/class.ts
feng-eric: That's definitely a good point. What's a use case you can see for soft delete? Perhaps this is something we can bring up with designers?

models/Class.ts
feng-eric: @mahadzaryab1 Is this required actually?

pages/api/class/[id].ts
Comment (cindy-wang328) [reply pages/api/class/[id].ts:37]: why does the next file have post and put but this one has delete :O

models/Class.ts
mahadzaryab1: yeah - the fields that are optional should have a `?`. It then sets them to undefined if a value isn't provided.

services/database/class.ts
mahadzaryab1: yeah I think types should be Pascal Case

mahadzaryab1: here's a good style guide - https://basarat.gitbook.io/typescript/styleguide

services/database/class.ts
mahadzaryab1: Soft delete just makes sure that we don't lose a row in the table forever. After something's been deleted for a while - we can just go in and hard delete them using a SQL script.

services/database/class.ts
feng-eric: @jason17h

services/database/class.ts
feng-eric: Yeah, but do you see a necessary use case for soft delete? Like where would it be useful to have a soft delete?

feng-eric: @mahadzaryab1 I'll merge it in for now. We can discuss with the entire team if soft delete is needed. If it is, we can make that change in the future.

PR #31, opened 2021-07-03
Add Waitlist Endpoints

feng-eric:
### Notion ticket link

[Waitlist Endpoints](https://www.notion.so/uwblueprintexecs/086c1b564dbf478faa649a8880e34b81?v=9d3f8002a0d1477ba395616dfc082af7&p=9da31aea03274a1b9fbab45ee9575544)

- GET endpoint for waitlists. If both parentId and classId are passed in, we find the specific waitlist record. If only the parentId is passed in, we get all waitlist records associated with the parent. If only the classId is passed in, we get all waitlist records associated with the class
- POST endpoint: Validate the input to ensure both parentId and classId are passed in and then create the waitlist record
- DELETE endpoint: Validate the input to ensure both parentId and classId are passed in and then delete the waitlist record 

### Steps to test

1. Try the endpoints on postman. In order to create a waitlist record, it has to be associated with a parent and class. Thus, those need to be created prior to creating a waitlist.

-  I haven't manually tested the code since it's blocked on parent and class stuff being complete so it would be best if you guys could look through it closely to minimize bugs.

### Checklist

-   [ ] I have run the linter

### DIFF
--- a/pages/api/waitlist/index.ts
+++ b/pages/api/waitlist/index.ts
@@ -60,7 +60,7 @@ export default async function handle(
                     res,
                     `Waitlist record could not be created`,
                 );
-                break;
+                return;
             }
             ResponseUtil.returnOK(res, newWaitlistRecord);
             break;
@@ -78,7 +78,7 @@ export default async function handle(
                     res,
                     `Waitlist record with parentId:${input.parentId}, classId:${input.classId} not found.`,
                 );
-                break;
+                return;
             }
             ResponseUtil.returnOK(res, deleteWaitlistRecord);
             break;
--- a/pages/api/waitlist/index.ts
+++ b/pages/api/waitlist/index.ts
@@ -1,0 +1,96 @@
+import { NextApiRequest, NextApiResponse } from "next";
+import { ResponseUtil } from "@utils/responseUtil";
+import {
+    getWaitlistRecord,
+    getWaitlistRecordsByClassId,
+    getWaitlistRecordsByParentId,
+    createWaitlistRecord,
+    deleteWaitlistRecord,
+} from "@database/waitlist";
+import { WaitlistInput } from "models/Waitlist";
+import { validateWaitlistRecord } from "@utils/validation/waitlist";
+
+/**
+ * handle controls the request made to the waitlist resource
+ * @param req API request object
+ * @param res API response object
+ */
+export default async function handle(
+    req: NextApiRequest,
+    res: NextApiResponse,
+): Promise<void> {
+    switch (req.method) {
+        // Retrieving a waitlist by passing in the classId and parentId.
+        case "GET": {
+            const input = req.body as WaitlistInput;
+
+            if (input.classId && input.parentId) {
+                // If both classId and parentId are passed in, retrieve the specific waitlist record
+                const waitlistRecord = await getWaitlistRecord(input);
+                ResponseUtil.returnOK(res, waitlistRecord);
+            } else if (input.classId) {
+                // If only classId was passed in, retrieve all waitlist records for the class
+                const waitlistRecordsForClass =
+                    await getWaitlistRecordsByClassId(input.classId);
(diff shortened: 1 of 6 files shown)


pages/api/waitlist/index.ts
```diff
@@ -0,0 +1,96 @@
...
+                ResponseUtil.returnBadRequest(res, validationErrors.join(", "));
+                return;
+            }
+            const newWaitlistRecord = await createWaitlistRecord(input);
+            if (!newWaitlistRecord) {
+                ResponseUtil.returnBadRequest(
+                    res,
+                    `Waitlist record could not be created`,
+                );
+                break;
```
mahadzaryab1: should this be a `return` instead?

pages/api/waitlist/index.ts
```diff
@@ -0,0 +1,96 @@
...
+                ResponseUtil.returnBadRequest(res, validationErrors.join(", "));
+                return;
+            }
+            const deletedWaitlistRecord = await deleteWaitlistRecord(input);
+            if (!deletedWaitlistRecord) {
+                ResponseUtil.returnNotFound(
+                    res,
+                    `Waitlist record with parentId:${input.parentId}, classId:${input.classId} not found.`,
+                );
+                break;
```
mahadzaryab1: should this be a `return` instead?

services/database/waitlist.ts
```diff
@@ -0,0 +1,97 @@
...
+import { WaitlistInput } from "models/Waitlist";
+/**
+ * getWaitlistRecord takes the classId and parentId and returns the associated waitlist
+ * @param classId
+ * @param parentId
+ */
+async function getWaitlistRecord(input: WaitlistInput): Promise<Waitlist> {
+    const waitlistRecord = await prisma.waitlist.findUnique({
+        where: {
+            parentId_classId: {
```
mahadzaryab1, line 12: what does this key `parentId_classId` do?

mahadzaryab1 (commented):
Added a few comments / questions. Also, can you make the PR title according to the conditions listed in the PR template?

pages/api/waitlist/index.ts
feng-eric: Good catch!

pages/api/waitlist/index.ts
feng-eric: Good catch!

services/database/waitlist.ts
feng-eric: Good question! When using findUnique, it has to be a primary key for the table. The primary key in this case is (parentId, classId). In Prisma, they will create a field specifically for the @@id specified of the model. See https://www.prisma.io/docs/reference/api-reference/prisma-client-reference/#findunique.

mahadzaryab1 (approved):
LGTM!

utils/validation/waitlist.ts
```diff
@@ -0,0 +1,20 @@
...
+
+    if (!input.classId) {
+        validationErrors.push("classId is not provided");
+    }
+
+    if (!input.parentId) {
+        validationErrors.push("parentId is not provided");
+    }
+
+    return validationErrors;
```
cindy-wang328, line 19: Do we need to check for duplicates, like what if the class id and parent id are already in the waitlist records?
Or what if the ids are not valid (like if they are not ints, or dont exist)? 
Sorry if this is already checked somewhere else😅 
Edit: Never mind, it is already handled in the previous files

feng-eric: I believe if the class id and parent id are already in the waitlist record, since the pair is the primary key of the table, prisma should throw an error. We could handle it before it reaches prisma but that would mean an additional query. Do you think we should change it to that? @cindy-wang328

feng-eric: @cindy-wang328 I'll merge this since it's not blocked on it but if you think we should do this, we can make that change in the future. If we do make the change, do you think we should check for existence of a user, a program, etc as well prior to creating?

PR #32, opened 2021-07-04
Impelement Update Program Endpoint

victoranyun:
### Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/Set-up-program-CRUD-endpoint-eb4f61da390f40b98d2321e90a5b3126)

- Added a new endpoint for PUT /api/program/:programId
- API is expecting the full body (exact same as what is passed when creating a program)
- Validating the new data similar to creating a new program record

Before update:
Screen Shot 2021-07-04 at 7 47 22 PM

After update:
Screen Shot 2021-07-04 at 7 49 30 PM

### Steps to test

1. Postman

- Exception handling
- Typing

### Checklist

-   [x] I have run the linter

### DIFF
--- a/models/Program.ts
+++ b/models/Program.ts
@@ -4,7 +4,7 @@ import type { programFormat } from "@prisma/client";
  * custom type used for the POST endpoint when creating a new program
  * id is not required and is automatically incremented in the record
  */
-export type CreateProgramInput = {
+export type ProgramInput = {
     price: number;
     onlineFormat: programFormat;
     tag: string;
--- a/pages/api/program/[id].ts
+++ b/pages/api/program/[id].ts
@@ -1,8 +1,8 @@
 import { NextApiRequest, NextApiResponse } from "next";
 import { ResponseUtil } from "@utils/responseUtil";
 import { getProgram, deleteProgram, updateProgram } from "@database/program";
-import { CreateProgramInput } from "models/Program";
-import { validateCreateProgram } from "@utils/validation/program";
+import { ProgramInput } from "models/Program";
+import { validateProgramData } from "@utils/validation/program";
 
 /**
  * handle takes the programId parameter and returns
@@ -48,9 +48,7 @@ export default async function handle(
         return;
     } else if (req.method == "PUT") {
         // validate new body
-        const validationError = validateCreateProgram(
-            req.body as CreateProgramInput,
-        );
+        const validationError = validateProgramData(req.body as ProgramInput);
         if (validationError.length !== 0) {
             ResponseUtil.returnBadRequest(res, validationError.join(", "));
             return;
@@ -60,7 +58,7 @@ export default async function handle(
         // obtain the entire update body
         const program = await updateProgram(
             id as string,
-            req.body as CreateProgramInput,
+            req.body as ProgramInput,
         );
 
         if (!program) {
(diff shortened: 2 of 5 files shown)


pages/api/program/[id].ts
```diff
@@ -35,6 +37,32 @@ export default async function handle(
...
+                res,
+                `Program with id ${id} not found.`,
+            );
+            return;
+        }
+        ResponseUtil.returnOK(res, program);
+        return;
+    } else if (req.method == "PUT") {
+        // validate new body
+        const validationError = validateCreateProgram(
```
Comment (feng-eric) [inline pages/api/program/[id].ts]: If the validation is same for create and update, let's rename the function name

pages/api/program/[id].ts
```diff
@@ -35,6 +37,32 @@ export default async function handle(
...
+            req.body as CreateProgramInput,
```
Comment (feng-eric) [inline pages/api/program/[id].ts]: Same as the name of the class

Comment (victoranyun) [reply pages/api/program/[id].ts]: Fixed.

pages/api/program/[id].ts
Comment (victoranyun) [reply pages/api/program/[id].ts]: Fixed.

models/Program.ts
```diff
@@ -4,7 +4,7 @@ import type { programFormat } from "@prisma/client";
  * custom type used for the POST endpoint when creating a new program
  * id is not required and is automatically incremented in the record
  */
-export type CreateProgramInput = {
```
mahadzaryab1, line 7: Update function documentation as this type isn't just used for `POST` anymore

services/database/program.ts
```diff
@@ -58,4 +56,18 @@ async function deleteProgram(id: string): Promise<Program> {
...
 }
-export { getProgram, getPrograms, createProgram, deleteProgram };
+
+async function updateProgram(
+    id: string,
+    updatedProgramData: ProgramInput,
+): Promise<Program> {
+    const program = await prisma.program.update({
+        where: {
+            id: parseInt(id),
```
mahadzaryab1, line 66: this probably doesn't matter in our case but what happens if id isn't a parseable int?

mahadzaryab1 (commented):
Added a few comments / questions

pages/api/program/[id].ts
```diff
@@ -35,6 +37,30 @@ export default async function handle(
...
+        if (!program) {
+            ResponseUtil.returnNotFound(
+                res,
+                `Program with id ${id} not found.`,
+            );
+            return;
+        }
+        ResponseUtil.returnOK(res, program);
+        return;
+    } else if (req.method == "PUT") {
```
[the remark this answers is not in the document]
Comment (feng-eric) [reply pages/api/program/[id].ts:49]: I think if we have more than 1 method in one file, let's change to switch statement to avoid else if statements

services/database/program.ts
feng-eric: Pretty sure it'll just return an error if it isn't an int

feng-eric: Worth testing though to verify @victoranyun

brandonwong368: If a float is passed in, parseInt rounds down and returns the integer, for example `parseInt(2.5)` gives you 2, the error is returned if that integer doesn't correspond to an existing record

Should we fix this in a separate PR that throws an error when floats are passed in?

feng-eric: We could also do this validation in our validation function. @brandonwong368  We can fix it in a different PR.

brandonwong368: Okay sounds good

brandonwong368 (approved):

brandonwong368:
@feng-eric @mahadzaryab1 should be good to merge then, both comments about parsingInt and switch/case are to be resolved in separate PRs

feng-eric (approved):
Good work @victoranyun @brandonwong368 !

PR #35, opened 2021-07-08
Validate user info

jason17h:
### Notion ticket link

[Validate User Info](https://www.notion.so/uwblueprintexecs/086c1b564dbf478faa649a8880e34b81?v=9d3f8002a0d1477ba395616dfc082af7&p=8cc69e4824c742b3bf7b6499b3ceddc6)

This PR adds a function to validate the data for the `User` and its corresponding role (`Parent`, `Volunteer`, `ProgramAdmin`) and calls it in the `updateUser` function before updating the user. If the user data is invalid, the function immediately returns nothing.

### Steps to test

1. Try making a `PUT` request for users and play around with valid/invalid fields.

-

### Checklist

-   [x] I have run the linter


utils/validation/user.ts
```diff
@@ -0,0 +1,108 @@
+import { locale, roles, province } from ".prisma/client";
+import { UserInput, ParentInput, VolunteerInput } from "models/User";
+import validator from "validator";
+
+/**
+ * Valdaites whether a given province has been specified correctly
+ * @param userProvince - the province to validate
+ * @returns - true if the province is valid, false otherwise
+ */
+function validateProvince(userProvince) {
```
feng-eric, line 34: Does the Validator library not have a province Validator?

utils/validation/user.ts
```diff
@@ -0,0 +1,108 @@
...
+    return [
+        province.AB,
```
feng-eric: Not sure if this would be better but perhaps we can make this a const outside the function

utils/validation/user.ts
```diff
@@ -0,0 +1,108 @@
...
+    ].includes(userProvince);
+}
+
+/**
+ * Validates whether the specified preferred language has been specified correctly
+ * @param userLanguage - the preferred language to validate
+ * @returns - true if the language is valid, false otherwise
+ */
+function validatePreferredLanguage(userLanguage) {
+    return [locale.en, locale.ja, locale.ko, locale.zh].includes(userLanguage);
```
feng-eric: Same as above comment

utils/validation/user.ts
feng-eric: Also, would it be better to use the Set object if we want to check for existence

feng-eric (commented):
Left some comments! Thanks for getting this out so quick btw 😄

utils/validation/user.ts
jason17h: No, and either way I think maybe it would be good to compare against the values we are using in our database just to ensure that it's consistent

feng-eric (approved):
LGTM! If you tested this manually, you can feel free to merge it in

victoranyun (approved):

PR #38, opened 2021-07-11
Export html to pdf

brandonwong368:
### Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/086c1b564dbf478faa649a8880e34b81?v=9d3f8002a0d1477ba395616dfc082af7&p=1be01de27e1c462bade88d67e612f917

- using html-pdf package
- placeholder variable for local path to html file or html string to convert into pdf
- downloads file to a designated path 

### Steps to test

1. Added a random html file (`test.html` match the filename to the util) to the repository root 
2. Created a new endpoint calling the function. Used postman to send a request
3. New pdf called `test.pdf` will be downloaded (the test html file and the pdf were deleted after testing)

-

### Checklist

-   [ ] I have run the linter


feng-eric:
@brandonwong368 Can you join the railway project if you haven't so we can do a PR deploy to be able to easily test this?

utils/htmlToPDF.ts
```diff
@@ -0,0 +1,23 @@
+import { create, CreateOptions } from "html-pdf";
+import { readFileSync } from "fs";
+
+/**
+ * converts hard coded html into a PDF and downloads the PDF locally
+ * @param {string} downloadPath download path of the generated pdf
+ */
+const htmlToPDF = (downloadPath: string): void => {
+    // path of the hard coded html or the html as a string
+    const htmlPath = "./test.html";
```
mahadzaryab1, line 10: @brandonwong368 @feng-eric how would we do this in a real use-case? are we going to have an HTML string within our code?

mahadzaryab1 (commented):
added a question

utils/htmlToPDF.ts
feng-eric: Yeah, we'll probably store a hard coded HTML string and fill it in with variables for user specific data

PR #41, opened 2021-07-16
Create API Endpoint For Parent Enrollments

mahadzaryab1:
### Notion ticket link

[Set up enroll to program endpoint](https://www.notion.so/uwblueprintexecs/Set-up-enroll-to-program-endpoint-891e6e3039af4180bfd1e59c9c5b1a97)

- Added a new API endpoint for when a parent enrols their child into a program
- Will do the second enrollment workflow in a separate PR

### Checklist

-   [x] I have run the linter

### DIFF
--- a/models/ParentRegistration.ts
+++ b/models/ParentRegistration.ts
@@ -1,0 +1,5 @@
+export type ParentRegistrationInput = {
+    parentId: number;
+    studentId: number;
+    classId: number;
+};
--- a/models/ParentRegistration.ts
+++ b/models/ParentRegistration.ts
@@ -1,5 +1,11 @@
+/**
+ * Request Body Input For Parent Enrollment
+ */
 export type ParentRegistrationInput = {
+    /** Unique identifier of the parent enrolling their child */
     parentId: number;
+    /** Unique identifier of the child being enrolled */
     studentId: number;
+    /** Unique identifier of the class the child is being enrolled in */
     classId: number;
 };
--- a/pages/api/enroll/child.ts
+++ b/pages/api/enroll/child.ts
@@ -1,17 +1,27 @@
 import { NextApiRequest, NextApiResponse } from "next";
 import { useRouter } from "next/router";
 import { ResponseUtil } from "@utils/responseUtil";
-import { getParentRegistration } from "@database/enroll";
+import {
+    createParentRegistration,
+    getParentRegistration,
+} from "@database/enroll";
+import { validateParentRegistrationRecord } from "@utils/validation/parentRegistration";
+import { ParentRegistrationInput } from "@models/ParentRegistration";
 
+/**
+ * handle controls the request made to the enroll/child resource.
+ * This endpoint is used when a parent is enrolling their child into a program
+ * @param req API request object
+ * @param res API response object
+ */
 export default async function handle(
     req: NextApiRequest,
     res: NextApiResponse,
 ): Promise<void> {
     switch (req.method) {
         case "GET": {
-            const router = useRouter();
             // obtain query parameters
-            const { studentId, classId } = router.query;
+            const { studentId, classId } = req.query;
 
             // verify that query parameters were passed in
             if (!studentId || !classId) {
@@ -26,7 +36,7 @@ export default async function handle(
             const classIdNumber = parseInt(studentId as string, 10);
 
             // verify that the query parameters were successfully converted to numbers
-            if (!studentId || !classId) {
+            if (isNaN(studentIdNumber) || isNaN(classIdNumber)) {
                 return ResponseUtil.returnBadRequest(
                     res,
                     "studentId and classId should be passed in as numbers",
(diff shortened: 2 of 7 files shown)


models/ParentRegistration.ts
```diff
@@ -0,0 +1,11 @@
+/**
```
feng-eric, line 1: Thoughts on naming file `Registration.ts` and having `VolunteerRegistrationInput` in here as well?

pages/api/enroll/child.ts
```diff
@@ -0,0 +1,97 @@
...
+            // verify that query parameters were passed in
+            if (!studentId || !classId) {
+                return ResponseUtil.returnBadRequest(
+                    res,
+                    "studentId and classId are required to obtain a program registration record",
+                );
+            }
+
+            // parse query parameters from strings to numbers
+            const studentIdNumber = parseInt(studentId as string, 10);
```
feng-eric, line 34: What happens if you don't specify a radix?

pages/api/enroll/child.ts
```diff
@@ -0,0 +1,97 @@
...
+            const classIdNumber = parseInt(studentId as string, 10);
+
+            // verify that the query parameters were successfully converted to numbers
+            if (isNaN(studentIdNumber) || isNaN(classIdNumber)) {
```
feng-eric, line 38: We should probably do this in existing endpoints too.

pages/api/enroll/child.ts
```diff
@@ -0,0 +1,97 @@
...
+            const parentRegistrationRecord = await getParentRegistration(
+                studentIdNumber,
+                classIdNumber,
+            );
+
+            // verify that the parent registration record could be obtained
+            if (!parentRegistrationRecord) {
+                return ResponseUtil.returnNotFound(
+                    res,
+                    "a registration entry was not found for the studentId and classId",
```
feng-eric, line 55:
```suggestion
                    "registration entry was not found for the studentId and classId",
```

pages/api/enroll/child.ts
```diff
@@ -0,0 +1,97 @@
...
+            );
+            if (!newRegistration) {
+                ResponseUtil.returnBadRequest(
+                    res,
+                    `Registration could not be created`,
+                );
+                return;
+            }
+
+            ResponseUtil.returnOK(res);
```
feng-eric, line 85:
```suggestion
            ResponseUtil.returnOK(res, newRegistration);
```

Thoughts on returning the created record in the response? I think that's what we have it in existing places.

services/database/enroll.ts
```diff
@@ -0,0 +1,48 @@
+import prisma from "@database";
+import { ParentRegistrationInput } from "models/ParentRegistration";
+import { ParentReg } from "@prisma/client";
+
+/**
+ * getParentRegistration obtains the registration record of a parent enrollment
+ *
+ * @param {number} studentId  unique identifier of the enrolled child
```
feng-eric, line 8:
```suggestion
 * @param {number} studentId unique identifier of the enrolled child
```

services/database/enroll.ts
```diff
@@ -0,0 +1,48 @@
...
+    });
+
+    return parentRegistrationRecord;
+}
+
+/**
+ * createParentRegistration creates a new registration record of a parent enrollment
+ *
+ * @param {ParentRegistrationInput} parentRegistrationData the data containing the details of the enrollment
+ * @returns {Promise<ParentReg>}
```
feng-eric, line 32:
```suggestion
 * @returns {Promise<ParentReg>} new registration record
```

feng-eric (commented):
Had few small comments but overall looks good. Were you able to test this?

pages/api/enroll/child.ts
```diff
@@ -0,0 +1,97 @@
...
+                req.body as ParentRegistrationInput,
+            );
+            if (!newRegistration) {
+                ResponseUtil.returnBadRequest(
+                    res,
+                    `Registration could not be created`,
+                );
+                return;
+            }
+
```
feng-eric, line 84: We need to email a confirmation to the user. Can you either add a TODO comment or call the sendEmail function here?

pages/api/enroll/child.ts
mahadzaryab1: According to [this ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt)- it doesn't default to base 10 so it's necessary to have it

pages/api/enroll/child.ts
mahadzaryab1: I tried doing this but because we don't `include` the fields when we create the objects, the response returned is empty.

models/ParentRegistration.ts
mahadzaryab1: Named it to `Enroll.ts` to be consistent with the resource naming

pages/api/enroll/child.ts
mahadzaryab1: I'll add it as a TODO because we can discuss if doing it in the same endpoint is the best way to go or if we should create another endpoint for it

feng-eric: I think we should have it here since I don't thin it's necessary to call from the frontend twice.

mahadzaryab1: cool - let's do it in a separate PR then

mahadzaryab1:
> Had few small comments but overall looks good. Were you able to test this?

Yeah - here's what a GET response looks

```
{
    "data": {
        "parentId": 1,
        "studentId": 1,
        "classId": 1,
        "createdAt": "2021-07-16T15:22:34.270Z",
        "updatedAt": "2021-07-16T15:22:34.269Z"
    }
}
```

feng-eric (approved):
LGTM! Good work 🚀

PR #33, opened 2021-07-05
make frontend home page

cindy-wang328:
screenshot on macbook air:
image
image

on big monitor:
image
image

other resizing
image
image

One problem I am stuck on is, right now if there is an element by itself on a row, it will take up the entire row. Pls help :( 
image
Also ignore index.tsx, that is just what I did to test, we shouldn't delete the existing stuff

PR #42, opened 2021-07-17
Revert "make frontend home page"

feng-eric:
Reverts uwblueprint/social-diversity-for-children#33


cindy-wang328 (approved):
OOF i didnt know those reverts would mess  it up

PR #47, opened 2021-07-21
Integrate Login Page And Add Relevant Redirections

mahadzaryab1:
### Notion ticket link

[Log in integration](https://www.notion.so/uwblueprintexecs/Log-in-integration-43780f8ddca34491befa6bdb1235a68b)

- Integrated the login page that was built with NextAuth.js

### Steps to test

~~1. Heading over to http://localhost:3000/ should redirect you to the login page if you're not already authenticated~~
2. If you're authenticated, then you'll be redirected to the root page with a message that you're authenticated

### Checklist

-   [x] I have run the linter

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -1,10 +1,4 @@
-import {
-    useSession,
-    signIn,
-    signOut,
-    getSession,
-    GetSessionOptions,
-} from "next-auth/client";
+import { useSession, signIn, signOut } from "next-auth/client";
 
 export default function Component() {
     const [session, loading] = useSession();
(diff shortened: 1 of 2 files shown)


jason17h:
Quick question - what is the expected behaviour for after I enter my email and click "Continue"? It currently says "An unexpected error occurred". Is that being implemented later or was that meant to be handled in this PR?

pages/index.tsx
```diff
@@ -17,3 +23,27 @@ export default function Component() {
...
+
+/**
+ * getServerSideProps runs before each page is rendered to check to see if a
+ * user has already been authenticated
+ */
+export async function getServerSideProps(context: GetSessionOptions) {
+    // obtain the next auth session
+    const session = await getSession(context);
+
+    // if the user is not authenticated redirect them to the login page
```
feng-eric: Should we be redirecting them to the login page or the home page?

feng-eric (commented):
Had one small question.

feng-eric:
> Quick question - what is the expected behaviour for after I enter my email and click "Continue"? It currently says "An unexpected error occurred". Is that being implemented later or was that meant to be handled in this PR?

I think @victoranyun is going to be working on this.

pages/index.tsx
jason17h: That's what I was wondering at first as well, but if our platform is specifically for the class management stuff then I would assume you need to log in to be able to access anything? Or are you allowed to browse programs even if you're not registered?

feng-eric: You're allowed to browse programs even if you're not logged in.

feng-eric: It's something we discussed in May/June during our team meetings

pages/login.tsx
```diff
@@ -6,12 +6,22 @@ import {
...
  * This is the page that a user will use to either login or register
  * to the SDC platform
  */
 export default function Login(): JSX.Element {
+    // hook to hold the user's email
+    const [email, setEmail] = useState("");
+
+    // signInWithEmail sends a login request to the user's email
+    const signInWithEmail = () => {
+        signIn("email", { email: email });
```
feng-eric: Thought's on making "email" a const?

feng-eric: Also, I think it can just be:
```suggestion
        signIn("email", { email });
```
Let me know if this works.

feng-eric (approved):

PR #52, opened 2021-07-25
implement login page high fidelity

brandonwong368:
### Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/086c1b564dbf478faa649a8880e34b81?v=9d3f8002a0d1477ba395616dfc082af7&p=e6a16dc2c0dd4c89adab2a8189da631e)

- updated the login page to the high fidelity design
- added back button component and added it to the verify page
-
image

image

### Steps to test

1.

-

### Checklist

-   [x] I have run the linter

### DIFF
--- a/pages/login.tsx
+++ b/pages/login.tsx
@@ -39,14 +39,19 @@ export default function Login(): JSX.Element {
             <Center h="500px">
                 <Box width="700px">
                     <Center>
-                        <Text fontWeight="700" fontSize="32px">
-                            Sign Up Or Sign In
+                        <Text fontWeight="700" fontSize="36px">
+                            Sign In
                         </Text>
                     </Center>
                     <Center>
-                        <Text fontWeight="400" fontSize="18px" mt="20px">
+                        <Text fontWeight="400" fontSize="16px" mt="20px">
                             Registration for Summer 2021 classes begin June 31,
                             2021
+                        </Text>
+                    </Center>
+                    <Center>
+                        <Text fontWeight="400" fontSize="16px" mt="50px">
+                            Email Address
                         </Text>
                     </Center>
                     <FormControl id="email">
@@ -54,8 +59,9 @@ export default function Login(): JSX.Element {
                             <Input
                                 width="366px"
                                 type="email"
-                                placeholder="Email address"
-                                mt="40px"
+                                placeholder="Enter email"
+                                textAlign="center"
+                                mt="5px"
                                 onChange={(e) => {
                                     if (isEmail(e.target.value)) {
                                         setEmail(e.target.value);
@@ -75,7 +81,7 @@ export default function Login(): JSX.Element {
                                 width="366px"
                                 fontSize="12px"
                                 fontWeight="400"
-                                mt="20px"
+                                mt="30px"
                                 _hover={{ bg: "#0C53A0" }}
                             >
                                 Please enter a valid email to continue.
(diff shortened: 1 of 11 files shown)


feng-eric (approved):
LGTM! Nice work

victoranyun (approved):

src/components/SDCWrapper.tsx
```diff
@@ -11,7 +11,9 @@ const SDCWrapper: React.FC<SDCWrapperProps> = (props): JSX.Element => {
     return (
         <Box minHeight={"100vh"} position={"relative"}>
             <Navbar session={props.session} />
-            <Box pb={DEFAULT_FOOTER_HEIGHT}>{props.children}</Box>
+            <Box pb={DEFAULT_FOOTER_HEIGHT} px="48">
```
jason17h, line 14: why is the `px=48` change needed / did you mean to change it for everything on the website, or is it a change that only needs to be applied to the login/verify pages?

jason17h: Actually I think this makes sense, but we need to be careful when merging this because some of the other components have `px=48` to align them with the footer/navbar, so once this gets merged I think it's gonna double the padding

jason17h (approved):
Looks really good! Just added 1 extra comment

src/components/BackButton.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { ArrowBackIcon } from "@chakra-ui/icons";
+import { HStack, Button, Link } from "@chakra-ui/react";
+import { useRouter } from "next/router";
+
+export const BackButton: React.FC = () => {
+    const router = useRouter();
+
+    return (
+        <HStack spacing={6}>
+            <Link onClick={() => router.back()}>
```
jason17h: do you think you can maybe add a prop called `onClick` or `handleClick` or something, and then for the `onClick` property here you would call `props.onClick` if it exists and default to `() => router.back()`? just to make this button more reusable (e.g., in the participant info form the back button just goes to the previous form page rather than the previous url, so I wouldn't be able to use this component as is)

Thanks for making this into a component though, I did not think to do that

src/components/BackButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+    onClick?: React.MouseEventHandler;
+};
+
+export const BackButton: React.FC<BackButtonProps> = (props) => {
+    const router = useRouter();
+    const onClick = props.onClick ? props.onClick : () => router.back();
+
+    return (
+        <HStack spacing={6}>
+            <Link onClick={onClick}>
```
jason17h: btw does this need to be wrapped in a `Link` tag or can we just move the `onClick` prop to the `Button` instead?

jason17h: same with `HStack`

src/components/SDCWrapper.tsx
feng-eric: Are those PRs already in staging?

pages/login.tsx
```diff
@@ -75,50 +81,55 @@ export default function Login(): JSX.Element {
...
                                 mt="20px"
-                                _hover={{ bg: "#0C53A0" }}
+                                _hover={{
+                                    bg: "#2C6AAD",
+                                }}
+                                borderRadius={100}
                                 onClick={() => {
                                     setValue(!value);
                                     setLocalStorageEmail(email);
                                     signIn("email", { email: email });
```
feng-eric:
```suggestion
                                    signIn("email", { email });
```

PR #54, opened 2021-07-28
add language as a primary key of programtranslation

cindy-wang328:
There needs to be a one to many relationship with Program and ProgramTranslation, prisma is making a 1 to 1 because we forgot to put the language as a primary key in the programtranslation, so we should add it in


feng-eric (approved):
LGTM!

mahadzaryab1 (approved):
LGTM

PR #56, opened 2021-07-31
Volunteer Enroll Endpoint and Database Updates

feng-eric:
### Notion ticket link

[Set up enroll to program endpoint for volunteers](https://www.notion.so/uwblueprintexecs/086c1b564dbf478faa649a8880e34b81?v=9d3f8002a0d1477ba395616dfc082af7&p=e1d062b407784f039e4c80c390296d42)

- Added volunteer enrollment endpoint
- Added checking session and role for both parent and volunteer enrollment
- Changed to obtain parentId/volunteerId from the session instead of from the request
- Updated parent_reg table to have (parentId, studentId, classId) as primary key
- Changed updated_at so that it's allowed to be null
- Removed coupon_users table as not required anymore @mahadzaryab1 

### Checklist

-   [x] I have run the linter


pages/api/enroll/child.ts
```diff
@@ -17,6 +19,22 @@ export default async function handle(
     req: NextApiRequest,
     res: NextApiResponse,
 ): Promise<void> {
+    const session = await getSession({ req });
+
+    // If there is no session or the user is not a parent
+    if (!session || session.role !== roles.PARENT) {
+        return ResponseUtil.returnUnauthorized(res, "Unauthorized");
```
mahadzaryab1: nit: could use a more descriptive error message - maybe something like `only users with the PARENT role can access this resource`

pages/api/enroll/child.ts
```diff
@@ -17,6 +19,22 @@ export default async function handle(
...
+    }
+
+    const parentId = session.id as number;
+    if (!parentId) {
+        // TODO: What should the error message be here?
```
mahadzaryab1: what you have LGTM - we control what's stored in the session so it should likely never get to this point anyway

pages/api/enroll/volunteer.ts
```diff
@@ -0,0 +1,117 @@
...
+ */
+export default async function handle(
+    req: NextApiRequest,
+    res: NextApiResponse,
+): Promise<void> {
+    const session = await getSession({ req });
+
+    // If there is no session or the user is not a volunteer
+    if (!session || session.role !== roles.VOLUNTEER) {
+        return ResponseUtil.returnUnauthorized(res, "Unauthorized");
```
mahadzaryab1: same comment as above

prisma/schema.prisma
```diff
@@ -29,9 +18,8 @@ model Parent {
   province          province
   preferredLanguage locale            @map("preferred_language")
   createdAt         DateTime          @default(now()) @map("created_at") @db.Timestamptz(6)
-  updatedAt         DateTime          @default(now()) @map("updated_at") @db.Timestamptz(6)
+  updatedAt         DateTime?         @map("updated_at") @db.Timestamptz(6)
```
mahadzaryab1, line 21: why do we feel like this field cant be defaulted to the current timestamp

utils/responseUtil.ts
```diff
@@ -26,6 +26,15 @@ export class ResponseUtil {
         res.status(400).json({ error: message });
     }
 
+    /**
+     * Method for returning a status 401 Bad Request response
```
mahadzaryab1: typo in comment - should be `401 Unauthorized`

mahadzaryab1 (approved):
Added a few comments/questions

pages/api/enroll/child.ts
feng-eric: good idea!

prisma/schema.prisma
feng-eric: I think there isn't really a need to set it to the default time since it's storing the same duplicate information as `createdAt`.

feng-eric: Also, it doesn't really make sure for the updatedAt time to be defaulted because it's technically the time of creation rather than the time it was updated.

utils/responseUtil.ts
feng-eric: Good catch!

pages/api/enroll/child.ts
feng-eric: Fixed!

mahadzaryab1 (approved):
LGTM

PR #57, opened 2021-08-01
Homepage and Class List Page Integration

cindy-wang328:
https://www.notion.so/uwblueprintexecs/086c1b564dbf478faa649a8880e34b81?v=9d3f8002a0d1477ba395616dfc082af7&p=b648ad37d4ef4f8981aa306adf601e8e

Steps to test

- `npx prisma studio` to open prisma studio. Created 2 program + programtranslations, 2 class + classtranslations, 1 user + teacher + teacherreg
- opened localhost 3000 and clicked the program cards, checked the "Loading" text appears when it's loading

image
image

image
image
image
image
image


feng-eric:
Can you rename the PR `Homepage and Class List Page Integration`

pages/api/cardinfo/class.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @param req API request object
+ * @param res API response object
+ */
+export default async function handle(
+    req: NextApiRequest,
+    res: NextApiResponse,
+): Promise<void> {
+    if (req.method == "GET") {
+        const { id } = req.query;
+        const result = await getClassInfoWithProgramId(id as string);
```
feng-eric: I think we should just create an endpoint in `api/class/index.ts` for this. We should return classes by programId in the backend and reutrn everything about each class to the frontend. We should let the frontend just filter/process the data to the way that's required.

pages/api/cardinfo/program.ts
```diff
@@ -0,0 +1,33 @@
...
+ * handle takes the programId parameter and returns
+ * the program associated with the programId
+ * @param req API request object
+ * @param res API response object
+ */
+export default async function handle(
+    req: NextApiRequest,
+    res: NextApiResponse,
+): Promise<void> {
+    if (req.method == "GET") {
```
feng-eric: I think this functionality is really similar to `api/program/[id].ts`. I think you should just use that and send everything to the frontend. Just let the frontend process the data in the format that it needs to be in.

services/database/programcardinfo.ts
```diff
@@ -0,0 +1,168 @@
+import prisma from "@database";
+import { ProgramCardInfo } from "models/Program";
+import { ClassCardInfo } from "models/Class";
+import { locale } from "@prisma/client";
+
+/**
+ * @param {string} id - programId
+ */
+async function getClassInfoWithProgramId(id: string): Promise<ClassCardInfo[]> {
```
feng-eric: Related to comment above, let's just have a generic getClassesByProgramId query that's used in `api/class`.

feng-eric (changes requested):
Left some comments. I think we should try to keep the backend endpoints generic by returning all the information to the frontend and let frontend decide how to process the data. If we try to be selective of what fields to return, I think we will reach a point where we'll have a lot of different endpoints that aren't really re-usable but are really similar. 

I think it's better to allow endpoints to return all information about a resource so that we can re-use the endpoints.

pages/api/cardinfo/program.ts
cindy-wang328: Where do we put the existing getProgram then? Does that mean we don't need any endpoint to get a Program object, every GET would get a ProgramCardInfo?

feng-eric: Like we discussed on Slack, I think we don't have to have an endpoint to get a Program object since we always want to get the program along with the language. Let me know if I'm mistaken

pages/index.tsx
```diff
@@ -28,7 +36,11 @@ export default function Component(props: ComponentProps): JSX.Element {
                 </Heading>
 
                 <Box>
-                    <ProgramList />
+                    {apiResponse ? (
+                        <ProgramList cardInfo={apiResponse.data} />
+                    ) : (
+                        "Loading"
```
mahadzaryab1: Can we use a [spinner](https://chakra-ui.com/docs/feedback/spinner) here?

pages/programdetails/[pid].tsx
```diff
@@ -1,153 +1,110 @@
...
+        fetchWithId,
+    );
+    const { data: programInfoResponse, error: programInfoError } = useSWR(
+        ["/api/cardinfo/program", pid],
+        fetchWithId,
+    );
+    if (classListError || programInfoError)
+        return (
+            <Text>
+                An error has occurred.{" "}
```
Comment (mahadzaryab1) [inline pages/programdetails/[pid].tsx]: what's the `.{" "}` for?

mahadzaryab1 (commented):
Added a few comments in addition to what @feng-eric said

pages/programdetails/[pid].tsx
Comment (cindy-wang328) [reply pages/programdetails/[pid].tsx]: LOL i have no idea, maybe i left an extra space somewhere and the formatter did this

cindy-wang328:
image
Spinner (the other spinner shows up too, but too fast to screenshot)

pages/api/class/index.ts
```diff
@@ -15,7 +16,11 @@ export default async function handle(
 ): Promise<void> {
     switch (req.method) {
         case "GET": {
-            const classes = await getClasses();
+            const { id } = req.query;
```
feng-eric: Let's name this programId so we don't get it confused with the id for class.

feng-eric:
```suggestion
            const { programId } = req.query;
```

pages/api/class/index.ts
```diff
@@ -15,7 +16,11 @@ export default async function handle(
...
+
+            const classes = id
```
feng-eric: Change to use programId here

pages/api/class/index.ts
```diff
@@ -15,7 +16,11 @@ export default async function handle(
...
+                ? await getClassInfoWithProgramId(id as string)
```
feng-eric: Let's convert `programId` to a number and use it as type number into this function. 

See below for example of converting to a number
https://github.com/uwblueprint/social-diversity-for-children/blob/862fabc0e0dd0709180aea47a5ddb30ae6e187d7/pages/api/enroll/child.ts#L43-L63

pages/api/program/[id].ts
```diff
@@ -14,69 +17,68 @@ export default async function handle(
...
-        if (!program) {
-            ResponseUtil.returnNotFound(
-                res,
-                `Program with id ${id} not found.`,
-            );
-            return;
+    switch (req.method) {
+        case "GET": {
+            const { id } = req.query;
+            const result = await getProgramCardInfo(id as string);
```
Comment (feng-eric) [inline pages/api/program/[id].ts]: Let's convert to a number and pass it in as a number instead.

See below on how to convert to a number here. Perhaps we could make this an util function? :O 

https://github.com/uwblueprint/social-diversity-for-children/blob/862fabc0e0dd0709180aea47a5ddb30ae6e187d7/pages/api/enroll/child.ts#L43-L63

pages/api/program/[id].ts
```diff
@@ -14,69 +17,68 @@ export default async function handle(
...
-    } else if (req.method == "DELETE") {
-        // Obtain program id
-        const { id } = req.query;
-        // Delete program with provided programId
-        const program = await deleteProgram(id as string);
+        case "DELETE": {
+            // Obtain program id
+            const { id } = req.query;
+            // Delete program with provided programId
+            const program = await deleteProgram(id as string);
```
Comment (feng-eric) [inline pages/api/program/[id].ts]: Same as above!

pages/programdetails/[pid].tsx
```diff
@@ -1,153 +1,126 @@
 import { useRouter } from "next/router";
```
Comment (feng-eric) [inline pages/programdetails/[pid].tsx]: Let's have the directory name be `pages/program-details/[pid].tsx`

services/database/programcardinfo.ts
```diff
@@ -0,0 +1,168 @@
+import prisma from "@database";
```
feng-eric: Let's try to keep naming consistent. I think `program-card-info.ts` would be better? What do you guys think? @mahadzaryab1 @cindy-wang328

services/database/programcardinfo.ts
```diff
@@ -0,0 +1,168 @@
...
+                    teacher: {
+                        select: {
+                            user: true,
+                        },
+                    },
+                },
+            },
+        },
+    });
+    return findResults.map((result) => {
```
feng-eric: What I remember doing last term is that we had util files which would do this mapping for converting the data into what we wanted. In database, we returned everything but we when used the function in the util file to clean it up a bit.

For example, in this case, we would first getClassByProgramId. Then, we would call this util function in the file in `api` to get the data in the format we want it to be. 

We could also do the same here, it would just be moving where you're mapping to a file within `util`. I think this would make the functions here a lot more usable if it returns all the information

services/database/programcardinfo.ts
```diff
@@ -0,0 +1,168 @@
...
+            imageLink: true,
+
+            programTranslation: {
+                where: {
+                    language: locale.en,
+                },
+            },
+        },
+    });
+    return findResult.map((p) => {
```
feng-eric: Same comment as above

feng-eric (commented):
Had some suggestions that I wanted to discuss. You don't have to change it if you don't agree with them

services/database/programcardinfo.ts
cindy-wang328: image
sure but LOL i was following these

services/database/programcardinfo.ts
cindy-wang328: The map doesn't really decrease amount of info, I included all the stuff in Program
The map is for flattening the name and description into the returned object, otherwise it'll have a nested programTranslation field
Program is 
```
  price INTEGER NOT NULL, -- price in cents, to make it integer
  online_format program_formats NOT NULL,
  tag TEXT NOT NULL, -- art, music, math etc (TODO we might need a tag table later)
  image_link TEXT,
  start_date TIMESTAMPTZ NOT NULL, -- earliest start date of all classes
  end_date TIMESTAMPTZ NOT NULL, -- latest end date of all classes
  is_archived BOOLEAN DEFAULT false NOT NULL,
  created_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
  updated_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP
```

pages/programdetails/[pid].tsx
Comment (mahadzaryab1) [reply pages/programdetails/[pid].tsx]: we can remove it then

services/database/programcardinfo.ts
feng-eric: I talked to @mahadzaryab1 and he said we should follow camelCase for typescript conventions actually

pages/program-details/[pid].tsx
```diff
@@ -41,14 +42,20 @@ export const ProgramDetails: React.FC = () => {
             </Text>
         );
     }
+    const mappedProgramInfoResponse = programInfoResponse
```
Comment (feng-eric) [inline pages/program-details/[pid].tsx]: Could name it `programCardInfo`?

pages/program-details/[pid].tsx
```diff
@@ -41,14 +42,20 @@ export const ProgramDetails: React.FC = () => {
...
+        ? CardInfoUtil.getProgramCardInfo(programInfoResponse.data)
+        : null;
+    const mappedClassInfoResponse = classListResponse
```
Comment (feng-eric) [inline pages/program-details/[pid].tsx]: Could name it `classCardInfo`?

utils/cardInfoUtil.ts
```diff
@@ -0,0 +1,89 @@
+import { ProgramCardInfo } from "models/Program";
+import { ClassCardInfo } from "models/Class";
+
+export class CardInfoUtil {
+    static getClassCardInfo(findResults: any[]): ClassCardInfo[] {
```
feng-eric: Let's make this plural `getClassCardInfos` maybe? I think having it plural would make it more reflective since we're passing and returning an array.

feng-eric: Let's add some documentation too please!

feng-eric (approved):

jason17h (approved):

PR #58, opened 2021-08-02
Add Functionality For Automatically Applying Coupons

mahadzaryab1:
### Notion ticket link

[[BACKEND] Allow Automatic Discounts To Be Applied In Stripe](https://www.notion.so/uwblueprintexecs/BACKEND-Allow-Automatic-Discounts-To-Be-Applied-In-Stripe-9d0c222464e944b7ba1f4b4e9791c3ac)

- Added functionality to automatically apply a coupon if a `couponId` is passed in

### Checklist

-   [x] I have run the linter

### DIFF
--- a/pages/api/checkout/session.ts
+++ b/pages/api/checkout/session.ts
@@ -30,29 +30,21 @@ export default async function sessionHandler(
                 allow_promotion_codes: true,
             };
 
-            // if a coupon ID is not passed in then allow promotion codes to be entered
-            if (!stripeRequest.couponId) {
-                const session = await stripe.checkout.sessions.create(
-                    stripeSession,
-                );
-
-                res.status(200).json({ sessionId: session.id });
-            }
-            // otherwise apply the discount from the coupon ID passed in
-            else {
+            // if a coupon ID is passed in then allow use it to apply a discount
+            if (stripeRequest.couponId) {
                 delete stripeSession["allow_promotion_codes"];
                 stripeSession["discounts"] = [
                     {
                         coupon: stripeRequest.couponId,
                     },
                 ];
-
-                const session = await stripe.checkout.sessions.create(
-                    stripeSession,
-                );
-
-                res.status(200).json({ sessionId: session.id });
             }
+
+            const session = await stripe.checkout.sessions.create(
+                stripeSession,
+            );
+
+            res.status(200).json({ sessionId: session.id });
             break;
         }
         default: {
--- a/pages/api/checkout/session.ts
+++ b/pages/api/checkout/session.ts
@@ -1,7 +1,7 @@
 import { StripeCheckoutRequest } from "@models/StripeCheckout";
 import { ResponseUtil } from "@utils/responseUtil";
 import { NextApiRequest, NextApiResponse } from "next";
-import { stripe } from "services/stripe";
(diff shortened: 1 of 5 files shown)


pages/api/checkout/session.ts
```diff
@@ -16,21 +16,46 @@ export default async function sessionHandler(
...
-                        quantity: stripeRequest.quantity,
-                    },
-                ],
-                mode: "payment",
-                success_url: `${req.headers.origin}/result?session_id={CHECKOUT_SESSION_ID}`,
-                cancel_url: `${req.headers.origin}/checkout`,
-                allow_promotion_codes: true,
-            });
+            // if a coupon ID is passed in then apply an automatic discount
+            if (stripeRequest.couponId) {
```
feng-eric: This is small but if we wanted to make the code cleaner, we could first use an object to store the data being created. Then, we use this if condition to conditionally set `discounts` in the object. After the else statement, we could have `const session = await stripe.checkout.sessions.create(checkoutObject)`. Just a suggestion to reduce the amount of duplicate/similar code! Don't need to change if you disagree.

feng-eric (approved):
Had small comment but other than that, LGTM!

pages/api/checkout/session.ts
mahadzaryab1: Great suggestion! Done.

pages/api/checkout/session.ts
```diff
@@ -28,9 +28,31 @@ export default async function sessionHandler(
                 success_url: `${req.headers.origin}/result?session_id={CHECKOUT_SESSION_ID}`,
                 cancel_url: `${req.headers.origin}/checkout`,
                 allow_promotion_codes: true,
-            });
+            };
 
-            res.status(200).json({ sessionId: session.id });
+            // if a coupon ID is not passed in then allow promotion codes to be entered
+            if (!stripeRequest.couponId) {
+                const session = await stripe.checkout.sessions.create(
```
feng-eric: You can move this part after the if else.

pages/api/checkout/session.ts
```diff
@@ -28,9 +28,31 @@ export default async function sessionHandler(
...
+            // otherwise apply the discount from the coupon ID passed in
+            else {
+                delete stripeSession["allow_promotion_codes"];
+                stripeSession["discounts"] = [
+                    {
+                        coupon: stripeRequest.couponId,
+                    },
+                ];
+
+                const session = await stripe.checkout.sessions.create(
```
feng-eric: Related to comment above! Since it's the same code, we can move it after the if else.

pages/api/checkout/session.ts
```diff
@@ -28,9 +28,31 @@ export default async function sessionHandler(
...
+                    stripeSession,
+                );
+
+                res.status(200).json({ sessionId: session.id });
+            }
```
feng-eric:
```suggestion
            }
            const session = await stripe.checkout.sessions.create(
                stripeSession,
            );
             
            res.status(200).json({ sessionId: session.id });
```

pages/api/checkout/session.ts
feng-eric:
```suggestion
```

PR #59, opened 2021-08-02
Parent Enrolment Select Child Page

brandonwong368:
### Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/086c1b564dbf478faa649a8880e34b81?v=9d3f8002a0d1477ba395616dfc082af7&p=fb6991c21a1b4af883fab2c7eaba0f27

- front end for the parent program registration/enrolment workflow
- added a group of radio buttons for each child associated with the parent given as an array (for now)
- only one child can be selected, the next button is disabled until a child is selected
image
image

### Steps to test

1.

- for the X button, it is slightly too far right, for the margin left, I tried to do calc(100%-48px) (48px is the default wrapper padding) to have it line up with the navbar, but couldn't get it to work, need a bit of help with this :(

### Checklist

-   [x] I have run the linter


pages/parentenrollclass.tsx
```diff
@@ -0,0 +1,142 @@
...
+    Center,
+    Text,
+    useRadio,
+    useRadioGroup,
+} from "@chakra-ui/react";
+import React, { useState } from "react";
+import { CloseIcon } from "@chakra-ui/icons";
+
+export default function ParentEnrollClass(): JSX.Element {
+    // Started creating buttons, found a thread that used radio buttons instead
```
feng-eric: What's this for?

pages/parentenrollclass.tsx
```diff
@@ -0,0 +1,142 @@
+import Wrapper from "@components/SDCWrapper";
```
feng-eric: Let's name the file `parent-enrollment.tsx`

pages/parentenrollclass.tsx
```diff
@@ -0,0 +1,142 @@
...
+        // children to be retrieved from parent enroll endpoint
+        const children = ["Christina Ru", "Raewyn Tsai", "Stacy Kwok"];
+
+        const { getRootProps, getRadioProps } = useRadioGroup({
+            name: "framework",
+            defaultValue: "react",
+            onChange: () => setIsDisabled(false),
+        });
+
+        const group = getRootProps();
```
feng-eric: What's getRootProps() doing?

pages/parentenrollclass.tsx
```diff
@@ -0,0 +1,142 @@
...
+                    height="50px"
+                    width="340px"
+                    borderRadius="6px"
+                    background="#737373"
+                    fontWeight="normal"
+                    textColor="#FFFFFF"
+                    fontSize="16px"
+                    isDisabled={isDisabled}
+                >
+                    {/* Link to the next page in the flow (does not exist yet) */}
```
feng-eric: Will the other steps of parent enrollment be in this file as well?

pages/parentenrollclass.tsx
```diff
@@ -0,0 +1,142 @@
...
+                    {props.children}
+                </Box>
+            </Box>
+        );
+    }
+    function ChildrenButtons() {
+        // children to be retrieved from parent enroll endpoint
+        const children = ["Christina Ru", "Raewyn Tsai", "Stacy Kwok"];
+
+        const { getRootProps, getRadioProps } = useRadioGroup({
```
feng-eric: Can you add some comments?

pages/parentenrollclass.tsx
```diff
@@ -0,0 +1,142 @@
...
+            name: "framework",
+            defaultValue: "react",
```
feng-eric: What's name and defaultValue?

feng-eric:
Let's also name the PR `Parent Enrollment Select Child Page`

pages/parentenrollclass.tsx
brandonwong368: No, I talked to Mahad/Stacy about this when the ticket was assigned and he said it would just be this page - the other pages would be on a separate task

pages/parentenrollclass.tsx
brandonwong368: I'm not exactly sure, I took some code from the chakra ui docs but I think it gives the corresponding component some default props, I console logged the `group` prop getter for the `Stack` component and took a screenshot
image

pages/parentenrollclass.tsx
feng-eric: Right, but I'm asking if it will be in this file as well, not if it's in the same task

pages/parent-enrollment.tsx
```diff
@@ -0,0 +1,125 @@
...
+                        </RadioCard>
+                    );
+                })}
+            </Stack>
+        );
+    }
+
+    return (
+        <Wrapper>
+            <Link href="/" ml={"100%"}>
```
jason17h: Re: your question in the PR description - Instead of doing `ml={"100%"}` in the link, try wrapping it in a `Flex` with the `alignItems` property set to `"right"`

src/components/CloseButton.tsx
```diff
@@ -1,19 +1,14 @@
...
 
-type CloseButtonProps = {
-    stypeProps?: Record<string, unknown>;
-    onClick?: React.MouseEventHandler;
-};
-
-export const CloseButton: React.FC<CloseButtonProps> = (props) => {
+export const CloseButton: React.FC<LinkProps> = (props) => {
     const router = useRouter();
     const onClick = props.onClick ? props.onClick : () => router.back();
```
jason17h, line 8: Thanks for fixing this for me! We can discuss whether it makes more sense to include a default behaviour or if we should just leave onClick as null. I realized that this line might be problematic if we don't want the close button to do anything on click (originally I thought it would be fine since you can override this to whatever you want your `onClick` to be, but I realized you cannot override it to null or else it will automatically be set to `router.back()`

pages/parentenrollclass.tsx
brandonwong368: I am really sorry, I misunderstood, I think the other pages will be in this file as well

pages/parent-enrollment.tsx
brandonwong368: :( the `alignItems` didn't work, but I did `justifyContent = "flex-end"` instead and it seemed to work

src/components/CloseButton.tsx
brandonwong368: yeah i agree, i think it's okay to keep the going back as default, it's mostly used as a back button substitute and you can easily pass href as a prop to change the link

feng-eric (commented):
I did notice one bug where the blue highlight won't appear until the second click.

You can see in the screenshot that the Next button is enabled but there isn't a blue outline
image

After second click:
image

feng-eric (approved):

pages/parent-enrollment.tsx
```diff
@@ -0,0 +1,121 @@
...
+            >
+                {props.children}
+            </Box>
+        </Box>
+    );
+}
+
+export default function ParentEnrollClass(): JSX.Element {
+    // Next button is disabled by default, activates when a child is selected
+    const [selectedChild, setSelectedChild] = useState<string>("");
```
jason17h, line 10: RE: this @feng-eric @brandonwong368 
> I did notice one bug where the blue highlight won't appear until the second click.
> 
> You can see in the screenshot that the Next button is enabled but there isn't a blue outline
> image
> 
> After second click:
> image

Hmm. I don't really know the root cause, but I noticed that if you move this call to `useState` to within the `ChildrenButtons` component, the box highlighting works fine (however, getting the parent component's state to update as well will require a bit more work). I can try to look into this more later but at least it's a start??

jason17h: Update: couldn't figure it out so we switched to buttons lol

pages/parent-enrollment.tsx
```diff
@@ -0,0 +1,76 @@
...
+                    </b>
+                    (select one)
+                </Text>
+            </Center>
+
+            <Center>
+                <VStack spacing={5}>
+                    {children.map((childName) => {
+                        return (
+                            <Button
```
jason17h: Can you also remove the box shadow? (Not the dark blue border, but the light blue thicker shadow around it that appears after you click it) I think you'd need to change the `_focus` prop to not have a box shadow. Let me know if you have trouble with it
Screen Shot 2021-08-05 at 10 01 23 PM

pages/parent-enrollment.tsx
```diff
@@ -0,0 +1,76 @@
...
+                                height="50px"
+                                width="340px"
+                                key={childName}
+                                onClick={() => setSelectedChild(childName)}
+                                border={
+                                    selectedChild === childName
+                                        ? "2px solid #0C53A0"
+                                        : "2px solid #E1E1E1"
+                                }
+                            >
```
jason17h: Also, I don't know if we want the box to go grey when you hover and click on it, maybe @christinaruu or @stacykwok can confirm this? I assume not though, in which case you would need to update the `_hover` and `_active` props to maintain the white background.
Screen Shot 2021-08-05 at 10 02 57 PM

pages/parent-enrollment.tsx
```diff
@@ -0,0 +1,76 @@
+import Wrapper from "@components/SDCWrapper";
+import { Flex, Button, Center, Text, VStack } from "@chakra-ui/react";
+import React, { useState } from "react";
+import { CloseButton } from "@components/CloseButton";
+
+export default function ParentEnrollClass(): JSX.Element {
+    // Next button is disabled by default, activates when a child is selected
+    const children = ["Christina Ru", "Raewyn Tsai", "Stacy Kwok"];
```
jason17h, line 9: maybe add a comment to clarify that this is just test data to be replaced during the integration with the backend?

pages/parent-enrollment.tsx
```diff
@@ -0,0 +1,76 @@
...
+                <CloseButton href="/"></CloseButton>
+            </Flex>
+            <Center>
+                <Text align="center" mt="15px" fontWeight="700" fontSize="36px">
+                    Program Registration
+                </Text>
+            </Center>
+            <Center>
+                <Text pb="55px" align="center" mt="30px">
+                    Who would you like to register for{" "}
```
jason17h, line 25: I think the `{" "}` might be there by mistake?

pages/parent-enrollment.tsx
```diff
@@ -0,0 +1,76 @@
...
+                    <b>
+                        Building Bridges with Music - <br />
+                        Singing Monkeys (Ages 9 and under)?{" "}
```
jason17h, line 28: same here about the `{" "}`

pages/parent-enrollment.tsx
```diff
@@ -0,0 +1,76 @@
...
+export default function ParentEnrollClass(): JSX.Element {
+    // Next button is disabled by default, activates when a child is selected
+    const children = ["Christina Ru", "Raewyn Tsai", "Stacy Kwok"];
+    const [selectedChild, setSelectedChild] = useState<string>("");
+
+    return (
+        <Wrapper>
+            <Flex justifyContent="flex-end">
+                {/* navigate to browse programs page instead of going back */}
+                <CloseButton href="/"></CloseButton>
```
jason17h:
```suggestion
                <CloseButton href="/" />
```

pages/parent-enrollment.tsx
```diff
@@ -0,0 +1,76 @@
...
+                        Singing Monkeys (Ages 9 and under)?{" "}
+                    </b>
+                    (select one)
+                </Text>
+            </Center>
+
+            <Center>
+                <VStack spacing={5}>
+                    {children.map((childName) => {
+                        return (
```
jason17h: small stylistic comment but since the function passed into `map` doesn't do anything but return something, you don't need to do 
```
(childName) => { 
    return (
        <Button>...</Button>
    )
 }
```
but can rather just do 
```
(childName) => (
    <Button>...</Button>
)
```
i.e. omit the `return` . you don't have to change this necessarily but just an FYI

jason17h (commented):
Looks good so far with the button changes, left some comments

pages/parent-enrollment.tsx
brandonwong368: I think this makes sure there is a space between the for and Building Bridges because of the bold tag, if not it would be like register for**Building Bridges**

pages/parent-enrollment.tsx
brandonwong368: @christinaruu @stacykwok lmk what you think, I thought it was worth trying this out, since it's a pretty easy fix/removal

pages/parent-enrollment.tsx
```diff
@@ -0,0 +1,76 @@
...
+                            fontWeight="normal"
+                            textColor="#0C53A0"
+                            borderRadius="6px"
+                            height="50px"
+                            width="340px"
+                            key={childName}
+                            onClick={() => setSelectedChild(childName)}
+                            border={
+                                selectedChild === childName
+                                    ? "2px solid #0C53A0"
```
feng-eric: This can be done as part of refactoring but since we're using the same colours in a lot of places, we should put them as variables to use. Just a note, don't need to make that change in this PR.

pages/parent-enrollment.tsx
christinaruu: yeah we think it would be better to differentiate the hover and clicked states, thanks for bringing it up! sorry about the changes :( 
here are the final states: https://www.figma.com/file/ad7RNFErTGP7HkL52edTwQ/SDC-External-Application?node-id=1574%3A18497

brandonwong368: hi @christinaruu , here are the new button states: 
Hover
image

Clicked/Selected
image

feng-eric (approved):

PR #73, opened 2021-09-13
Added firstTimeSetup script

gregmaxin:
A quick PR to help with the setup of the dev environment. This is for Ubuntu and WSL devs

Note: this does not cover 

- [ ] Setting up Heroku (you should follow [this](https://www.notion.so/uwblueprintexecs/Local-Database-Setup-82d0c72f8361447a95394967522477f3)
- [x] Setting up the environment file

### DIFF
--- a/fireTimeSetup.sh
+++ b/fireTimeSetup.sh
@@ -1,0 +1,10 @@
+sudo apt-get update
+sudo apt-get upgrade
+
+sudo apt-get install nodejs node.js npm
+
+#heroku
+curl https://cli-assets.heroku.com/install.sh | sh
+
+npm install
+npm update
--- a/fireTimeSetup.sh
+++ b/fireTimeSetup.sh
@@ -6,5 +6,4 @@ sudo apt-get install nodejs yarnpkg npm
 #heroku
 curl https://cli-assets.heroku.com/install.sh | sh
 
-npm install
-npm update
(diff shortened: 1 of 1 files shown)


fireTimeSetup.sh
```diff
@@ -0,0 +1,10 @@
+sudo apt-get update
+sudo apt-get upgrade
+
+sudo apt-get install nodejs node.js npm
+
+#heroku
+curl https://cli-assets.heroku.com/install.sh | sh
+
+npm install
+npm update
```
MatoPlus: We are using yarn, our dependency lock is in our `yarn.lock` file. We can replace it with a yarn install.
```suggestion
yarn
```

fireTimeSetup.sh
```diff
@@ -0,0 +1,10 @@
+sudo apt-get update
+sudo apt-get upgrade
+
+sudo apt-get install nodejs node.js npm
```
MatoPlus: We also want to install `yarn`, is `node.js` a real package?
```suggestion
sudo apt-get install nodejs yarnpkg npm
```

MatoPlus (commented):
Looking good! Thanks for setting this up. Just a few concerns on my end.

MatoPlus (approved):
LGTM!

PR #79, opened 2021-09-30
feature - 404 error page

aamyli:
### Notion ticket link

[Create 404 Error Page](https://www.notion.so/uwblueprintexecs/Create-404-Error-Page-2c23078e84b1404b9ea288b477764746)

(updated 10/03, this is what it looks like after adding suggestions)
Screen Shot 2021-10-03 at 5 48 25 PM

Error page designed according to the [Figma design here](https://www.figma.com/file/ad7RNFErTGP7HkL52edTwQ/SDC-%2F-S21-External-Application?node-id=1563%3A18312)

- Added `404.tsx` page and also the error icon as a component (`SvgErrorIcon`) as taken from the Figma

### Steps to test

1. Go to an invalid URL
2. Click the button to go back to home

- Let me know if there are any thoughts on styling fixes !

### Checklist

-   [x] I have run the linter

### DIFF
--- a/pages/404.tsx
+++ b/pages/404.tsx
@@ -13,13 +13,13 @@ export default function Custom404(props: ComponentProps): JSX.Element {
 
     return (
         <Wrapper session={props.session}>
-            <Center h="80vh" align="center">
-                <Box align="center" width="60%">
+            <Center minHeight="85vh" align="center">
+                <Box align="center" width="70%">
                     <SvgErrorIcon />
-                    <Heading size="xl" margin="3% 0% 3% 0%">
+                    <Heading size="md" margin="3% 0% 3% 0%">
                         Oh no! Page not found.
                     </Heading>
-                    <Text size="2xl" marginBottom="3%">
+                    <Text size="md" marginBottom="3%">
                         Sorry, but the page you are looking for does not exist.
                         Try refreshing the page or hit the button below.
                     </Text>
@@ -27,8 +27,8 @@ export default function Custom404(props: ComponentProps): JSX.Element {
                         color="white"
                         backgroundColor="#0C53A0"
                         _hover={{ backgroundColor: "#2C6AAD" }}
-                        width="50%"
-                        padding="10px 12px 10px 12px"
+                        size="sm"
+                        padding="10px 40px 10px 40px"
                         onClick={goToHome}
                         borderRadius={100}
                         fontWeight={"200"}
--- a/pages/404.tsx
+++ b/pages/404.tsx
@@ -1,5 +1,6 @@
 import Wrapper from "@components/SDCWrapper";
 import SvgErrorIcon from "@components/icons/ErrorIcon";
+import colourTheme from "@styles/colours";
 import { Button, Box, Center, Heading, Text } from "@chakra-ui/react";
 
 type ComponentProps = {
@@ -16,19 +17,22 @@ export default function Custom404(props: ComponentProps): JSX.Element {
             <Center minHeight="85vh" align="center">
                 <Box align="center" width="70%">
                     <SvgErrorIcon />
-                    <Heading size="md" margin="3% 0% 3% 0%">
+                    <Heading size="lg" mt={12}>
                         Oh no! Page not found.
                     </Heading>
-                    <Text size="md" marginBottom="3%">
+                    <Text size="md" my={9}>
                         Sorry, but the page you are looking for does not exist.
                         Try refreshing the page or hit the button below.
                     </Text>
                     <Button
                         color="white"
-                        backgroundColor="#0C53A0"
-                        _hover={{ backgroundColor: "#2C6AAD" }}
+                        backgroundColor={colourTheme.colors.Blue}
+                        _hover={{
+                            backgroundColor: colourTheme.colors.LightBlue,
+                        }}
                         size="sm"
-                        padding="10px 40px 10px 40px"
+                        py={5}
+                        width="50%"
                         onClick={goToHome}
                         borderRadius={100}
                         fontWeight={"200"}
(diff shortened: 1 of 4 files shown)


pages/404.tsx
```diff
@@ -0,0 +1,42 @@
...
+    const goToHome = () => {
+        window.location.href = "/";
+    };
+
+    return (
+        <Wrapper session={props.session}>
+            <Center minHeight="85vh" align="center">
+                <Box align="center" width="70%">
+                    <SvgErrorIcon />
+                    <Heading size="md" margin="3% 0% 3% 0%">
```
MatoPlus: little chakra-ism, `mt` is the same as margin for margin top. We also like to use unitless ticks (1, 2, 3, 4, ... 12), especially when we are just eyeballing. Heading size could be bigger!
```suggestion
                    <Heading size="lg" mt={12}>
```

pages/404.tsx
```diff
@@ -0,0 +1,42 @@
...
+                        Oh no! Page not found.
+                    </Heading>
+                    <Text size="md" marginBottom="3%">
```
MatoPlus: Chakra shorthand + beefier padding
```suggestion
                    <Text size="md" my={9}>
```

pages/404.tsx
```diff
@@ -0,0 +1,42 @@
...
+                        Sorry, but the page you are looking for does not exist.
+                        Try refreshing the page or hit the button below.
+                    </Text>
+                    <Button
+                        color="white"
+                        backgroundColor="#0C53A0"
```
MatoPlus: Please `import colourTheme from "@styles/colours";` and get colours from a source of truth
```suggestion
                        backgroundColor={colourTheme.colors.Blue}
```

pages/404.tsx
```diff
@@ -0,0 +1,42 @@
...
+                        _hover={{ backgroundColor: "#2C6AAD" }}
```
MatoPlus:
```suggestion
                        _hover={{
                            backgroundColor: colourTheme.colors.LightBlue,
                        }}
```

pages/404.tsx
```diff
@@ -0,0 +1,42 @@
...
+                        size="sm"
+                        padding="10px 40px 10px 40px"
```
MatoPlus, line 31: I don't think we want any padding in Chakra buttons. Default buttons are really responsive!

```suggestion
 
```

src/components/icons/ErrorIcon.tsx
```diff
@@ -0,0 +1,39 @@
...
+                y="97.428"
+                width="11.0149"
+                height="11.0149"
+                rx="5.50744"
+                fill="white"
+            />
+            <path
+                d="M55.3574 124C55.3574 124 65.6908 117.357 78.6074 117.357C91.5241 117.357 101.857 124 101.857 124"
+                stroke="white"
+                stroke-width="7"
```
MatoPlus: Pro chakra tip: every prop is camelCase. This would not be a valid property, the console should be complaining 😉 
```suggestion
                strokeWidth="7"
```

src/components/icons/ErrorIcon.tsx
```diff
@@ -0,0 +1,39 @@
...
+                stroke-linecap="round"
```
MatoPlus:
```suggestion
                strokeLinecap="round"
```

MatoPlus (changes requested):
Great stuff Amy! Just little code style suggestions from me! Along with some chakra css tips and we can ship this!

pages/404.tsx
MatoPlus: I think using a width % will be more responsive here. Also use beefier tick paddings.

```suggestion
                        py={5}
                        width="75%"
```

MatoPlus (changes requested):

MatoPlus (approved):
LGTM! Thanks for getting this done despite the busy week.

bekahma (commented):
@vedantdsn and I noticed that the button border radius is supposed to 6px (more rectangular). other than that, everything looks great!

image

MatoPlus:
> @vedantdsn and I noticed that the button border radius is supposed to 6px (more rectangular). other than that, everything looks great!
> 
> image

@bekahma  Amy actually brought up a really good point about this. She noticed that since most of our other buttons had round border radius unlike their design figma counterparts, she kinda thought that it would make sense to continue that and go with a round border instead.

Figma:
image
Actual:
image


I don't any strong feelings here, I'm fine with either. I'll let @bekahma @vedantdsn take the final call.

bekahma:
at the design handoff, the previous designer did mention that they made the switch from round to square buttons. i assume it was because they wanted to better differentiate tags from buttons ("Online" and "Music" in screenshot).

Screen Shot 2021-10-04 at 10 47 44 AM

i agree with @aamyli and it would be nice to keep the consistency so i'm okay with keeping it round only if some pages already have that round design (but that would mean future screens should have the round design as well). however, if time allows i would prefer converting round buttons to square. @MatoPlus really your call here for logistics and if it's possible to maybe create a ticket at the end of the external platform MVP as a final check for all button border widths across the app.

MatoPlus:
> at the design handoff, the previous designer did mention that they made the switch from round to square buttons. i assume it was because they wanted to better differentiate tags from buttons ("Online" and "Music" in screenshot).
> 
> Screen Shot 2021-10-04 at 10 47 44 AM
> 
> i agree with @aamyli and it would be nice to keep the consistency so i'm okay with keeping it round only if some pages already have that round design (but that would mean future screens should have the round design as well). however, if time allows i would prefer converting round buttons to square. @MatoPlus really your call here for logistics and if it's possible to maybe create a ticket at the end of the external platform MVP as a final check for all button border widths across the app.

Sounds great to me! I'll create another ticket to deal with all of our round borders. Stamped!

PR #80, opened 2021-09-30
refactor parent/volunteer signup

brandonwong368:
### Notion ticket link

[Refactor Participant/Volunteer Information Page](https://www.notion.so/uwblueprintexecs/086c1b564dbf478faa649a8880e34b81?v=9d3f8002a0d1477ba395616dfc082af7&p=6f6ab92a15ee4fbab36fc17d31f76059)

### Implementation description

- separated each page of the parent/volunteer signup into separate components
- eg. ParticipantInfoPage, EmergInfoPage, etc. under components
- replaced hardcoded colour values with colourTheme from @styles/colours

### Steps to test

1. Go to localhost:3000/parent/signup and localhost:3000/volunteer/signup
2. Complete the registration, pre/post refactor should have the same functionality

### What should reviewers focus on?

- Dev: functionality of overall flow
- 

### Checklist

-   [x] My PR name is descriptive and in imperative tense
-   [x] I have run the linter
-   [x] I have requested a review from the PL, and/or other devs who have background knowledge on this PR or who will be building on top of this PR

PR #83, opened 2021-10-03
Everything Signup

gregmaxin:
### Notion ticket link

Completes the following Tickets (12 Total Points):

- [x] [Form Validation of Inputs for Participant Info when Registering](https://www.notion.so/uwblueprintexecs/Form-Validation-of-Inputs-for-Participant-Info-when-Registering-28c120718cb7463db3deb45b135b8fb4)
- [x] [Registration Auto-Save](https://www.notion.so/uwblueprintexecs/Registration-Auto-Save-f759fcf4f76f4aaea1425e7235b204b1)
- [x] [Integrate User Setup Flow](https://www.notion.so/uwblueprintexecs/Integrate-User-Setup-Flow-2843b5475f9f4cf1b7f83624586eb453)
- [x] [Participant Form Regression](https://www.notion.so/uwblueprintexecs/Participant-Form-Regression-abbdfe7939b944f1ac488e08f2673515)

- [x] Modularized the field types to be their own components. These can be found under `src\components\formFields`. The goal is to hopefully make the creation of forms in the future much easier
- [x] Updated Both the Parent and Volunteer Signup flow to use the modularized components
- [x] The new modularized fields handle validations and are required as necessary 
- [x] Each Page in the signup flow prevents advancement to the next page if the validations or required fields are not completed correctly
- [x] Added the `Personal Info` and `Health Info` pages missing in the parent signup 
- [x] Fixed the states so that they now ALL properly map to the database
- [x] All form fields now auto-save using localstorage
- [x] Fixed the API calls so the data is passed to the backend
- [x] Added a spinner during account creation after the signup flow and an error message on failure

TODO (need help):
- [ ] The parent signup flow keeps failing with an `error 400` bad request. It passes the validation check-in in `pages\api\user\index.ts` but fails on line 43 of this file @ `const updatedUser = await updateUser(updatedUserData);` I'd appreciate if someone more familiar with the backend can look into this. 

### Steps to test

1. Test the flow on http://localhost:3000/volunteer/signup and http://localhost:3000/parent/signup

The error handling of phone numbers, dates and postal codes

-

### Checklist

-   [x] My PR name is descriptive and in the imperative tense
-   [x] I have run the linter


pages/parent/signup.tsx
```diff
@@ -421,55 +401,17 @@ export default function ParticipantInfo({
...
-        setHeardFromOptions([]);
-    };
-
+    //Form is finished
     async function updateUserAndClearForm() {
+        //Save the user
         const updatedUser = await updateUser();
         if (updatedUser) {
-            clearLocalStorage();
+            localStorage.clear();
```
gregmaxin, line 534: As mentioned in the discord, this may cause issues with the session auth. 

However, there must be a better way to do this than what we currently have (which actually misses some cases)

src/components/parent-form/EmergInfoPage.tsx
```diff
@@ -62,6 +81,22 @@ export const EmergInfoPage: React.FC<EmergPageProps> = ({
                     value={props.parentRelationship}
                 />
             </FormControl>
+            <div>
```
MatoPlus: When possible please try to try to use chakra counter parts, in this case, try using Box instead of div
```suggestion
            <Box>
```

pages/parent/signup.tsx
```diff
@@ -174,32 +181,6 @@ export default function ParticipantInfo({
...
-                <FormLabel>Please provide any details if necessary</FormLabel>
-                <Input
-                    placeholder="Details"
-                    onChange={(e) => setAllergies(e.target.value)}
-                    value={allergies}
-                />
-            </FormControl>
-        </Box>
-    ) : null;
-
```
MatoPlus: I think we actually want to keep these, it corresponds to the following form which has been removed, it seems like a regression from https://github.com/uwblueprint/social-diversity-for-children/pull/80. If you could, please ensure that the parent/signup flow matches exactly of that in our Figma.

image

But we have this currently for participant health form, which doesn't make sense:
image


And also do the required validations here.

pages/parent/signup.tsx
```diff
@@ -20,8 +20,6 @@ import { HeardFromPage } from "@components/parent-form/HeardFromPage";
 import { ParentCreatedPage } from "@components/parent-form/ParentCreatedPage";
 import colourTheme from "@styles/colours";
 
-const RADIO_YES = "yes";
-const RADIO_NO = "no";
```
MatoPlus, line 24: This will be used in the regression fix, please revert this delete.

src/components/parent-form/ParticipantInfoPage.tsx
```diff
@@ -51,44 +59,69 @@ export const ParticipantInfoPage: React.FC<ParticipantPageProps> = ({
...
+                            props.setParticipantLastName(e.target.value)
+                        }
+                        value={props.participantLastName}
+                    />
+                    <FormErrorMessage>{"Required"}</FormErrorMessage>
+                </FormControl>
+            </HStack>
+            <FormControl id="date-of-birth" isRequired>
+                <FormLabel>Date Of Birth</FormLabel>
+                <div
```
MatoPlus: Please use Box instead.

MatoPlus (changes requested):
Looking good Greg, I'll review this more later today but a couple of concerns ATM

Also, the alignment when the "required" prompt pops up really bug me, might wanna give the form more space so it doesn't jump around:
image

Edit: actually the above case is fine, I think it's probably something we can deal with for now.

src/components/parent-form/EmergInfoPage.tsx
gregmaxin: I wasn't aware that `<box/>` replaced `<div/>` I will use that from now on

pages/parent/signup.tsx
gregmaxin: Yes it was likely something from #80

pages/parent/signup.tsx
```diff
@@ -421,55 +401,17 @@ export default function ParticipantInfo({
             body: JSON.stringify(userData),
         };
         const response = await fetch("api/user", request);
+        console.log(response);
```
feng-eric:
```suggestion
```

utils/validation/fields.ts
```diff
@@ -0,0 +1,13 @@
+export function testPhoneNumber(phonenumber: string) {
```
feng-eric: Was there any particular reason why the [`validator` library ](https://www.npmjs.com/package/validator)wasn't used?

Here's a code snippet where we used it before:
https://github.com/uwblueprint/social-diversity-for-children/blob/dcdacb840ecb4de36370aa7490256dc23336156b/utils/validation/user.ts#L90-L100

feng-eric: If you feel that it isn't able to work, perhaps we should do our own regex validation in the code snippet too then instead of using `validator`

package.json
```diff
@@ -26,14 +26,17 @@
         "@prisma/client": "^2.23.0",
         "@stripe/stripe-js": "^1.15.0",
         "aws-sdk": "^2.926.0",
+        "formik": "^2.2.9",
```
feng-eric: Seems like this isn't currently being used. Will we be adding stuff from formik in this PR or will that be a different PR in the future?

pages/parent/signup.tsx
MatoPlus: I guess this is fine, but we would want to keep a note in case this cases a regression in the future.

utils/validation/fields.ts
MatoPlus: @GodGreg I feel like we should definitely use `validator` instead, especially if we want to make consistent use of the existing validation utils.

package.json
MatoPlus: From what you have currently, I think it's fine to just not use Formik for now for the sake of getting the MVP done. We can come back to it later for refinement. I'll create another ticket for this.

```suggestion
```

pages/parent/signup.tsx
MatoPlus: @GodGreg No one is currently working on fixing the regression, if you want to fix it, feel free to quickly add it in, there is no validation done for that page.

utils/validation/fields.ts
gregmaxin: I was unaware this existed; I will make the change

pages/parent/signup.tsx
```diff
@@ -421,55 +401,17 @@ export default function ParticipantInfo({
             body: JSON.stringify(userData),
         };
         const response = await fetch("api/user", request);
```
MatoPlus: The reason why this is doing 404 is because it's calling `/parent/user/user` on your page. We want do not want to do relative path:
```suggestion
        const response = await fetch("/api/user", request);
```

Similarly, please fix this for the `volunteer/signup` page!

After, I'm getting this error, so we definitely want to use validator's phone validate method for consistency: 
```json
{
"error": "Invalid phone number provided: "
}
```

pages/parent/signup.tsx
```diff
@@ -421,55 +401,17 @@ export default function ParticipantInfo({
             body: JSON.stringify(userData),
```
MatoPlus: Here, I think we want userData to use the emergency contact, since I don't actually think the username name columns get filled out initially when parents signup for an account, this is actually where they set their name. 

```
const userData = {
    firstName: emergFirstName,
    lastName: emergLastName,
    role: roles.PARENT,
    roleData: parentData,
};
```

My understanding is that we don't actually need these lines you had earlier:
```
    // Parent info
    const [parentFirstName, setParentFirstName] = useLocalStorage(
        "parentFirstName",
        "",
    );
    const [parentLastName, setParentLastName] = useLocalStorage(
        "parentLastName",
        "",
    );
    const [parentPhoneNumber, setParentPhoneNumber] = useLocalStorage(
        "parentPhoneNumber",
        "",
    );
    const [parentRelationship, setParentRelationship] = useLocalStorage(
        "parentRelationship",
        "",
    );
```

Edit: This might not actually be what we actually want, which you might have caught on, but this works a temporary solution until we find our how we can populate a new user name (maybe from their email account).

MatoPlus (changes requested):
Hey Greg, just a couple more stuff I picked up from your code.

I think we can probably make the name fields (emergency fields + volunteer name fields) required as they are expected by the api endpoint.

Please fix the regression if you can, it would be awesome if by the end of this ticket, we can confidently go through both signup forms, update the DB info with only the required data filled out. Great work so far 💪

bekahma (commented):
in terms of text field validations like postal code and phone number, everything looks good to me! @GodGreg i'm curious if there's a reason why a volunteer's name is not required? i know it's not explicitly stated in the designs but wondering what your thoughts are for that :) @MatoPlus @vedantdsn any thoughts?

Screen Shot 2021-10-05 at 8 38 44 PM

pages/parent/signup.tsx
gregmaxin: Turns out we don't need these with my implementation

package.json
gregmaxin: Removed

MatoPlus (commented):
As part of this ticket, could you also make sure that the `/signup` page fetches serverside session prop for the wrapper to be handled as well? Looks like I forgot to patch that in.

image
^ always "signin" button.

In fact, we actually want to redirect people to the login page if they do not have a session.

MatoPlus (commented):
I'll take a look at this later today, but as talked about in person, try to refine the fields as to what is actually required, fix the issue where difficulty/therapy is not saved.

You might also want to rebase origin/staging otherwise your current PR deploy will just be broken. For the merge conflict, the easiest way is probably to resolve the json and to regenerate the yarn.lock file again.

Thanks for cracking this out in one night, you're nuts! 💪

pages/parent/signup.tsx
```diff
@@ -90,20 +150,34 @@ export default function ParticipantInfo({
...
+        useLocalStorage("physical", false);
+    const [hasSensoryDifficulties, setHasSensoryDifficulties] = useLocalStorage(
+        "sensory",
+        false,
+    );
     const [participantDifficulties, setParticipantDifficulties] =
         useLocalStorage("participantDifficulties", []);
-    const [hasOtherDifficulties, setHasOtherDifficulties] = useState(false);
+    const [hasOtherDifficulties, setHasOtherDifficulties] = useLocalStorage(
+        "otherDifficulties",
```
MatoPlus: Conflicting names with the actual otherDifficulties storage
```suggestion
        "hasOtherDifficulties",
```

pages/parent/signup.tsx
```diff
@@ -112,16 +186,22 @@ export default function ParticipantInfo({
...
+    );
+    const [counseling, setCounseling] = useLocalStorage("conseling", false);
+    const [artTherapy, setArtTherapy] = useLocalStorage("art", false);
     const [participantTherapy, setParticipantTherapy] = useLocalStorage(
         "participantTherapy",
         [],
     );
-    const [hasOtherTherapy, setHasOtherTherapy] = useState(false);
+    const [hasOtherTherapy, setHasOtherTherapy] = useLocalStorage(
+        "otherTherapy",
```
MatoPlus: Conflicting name i think

```suggestion
        "hasOtherTherapy",
```

pages/parent/signup.tsx
MatoPlus: resolved, greg added the parent info form to flow

pages/volunteer/signup.tsx
```diff
@@ -183,36 +234,30 @@ export default function VolunteerInfo({
             headers: { "Content-Type": "application/json" },
             body: JSON.stringify(userData),
         };
-        const response = await fetch("api/user", request);
-        const updatedUserData = await response.json();
-        return updatedUserData;
+        const response = await fetch("/api/user", request);
+        if (response.status === 200) {
```
MatoPlus: I think this would be safer.
```suggestion
        if (response.ok) {
```

pages/parent/signup.tsx
```diff
@@ -420,65 +521,33 @@ export default function ParticipantInfo({
             headers: { "Content-Type": "application/json" },
             body: JSON.stringify(userData),
         };
-        const response = await fetch("api/user", request);
-        const updatedUserData = await response.json();
-        return updatedUserData;
+        const response = await fetch("/api/user", request);
+        if (response.status === 200) {
```
MatoPlus: safer?
```suggestion
        if (response.ok) {
```

utils/useLocalStorage.ts
```diff
@@ -7,7 +7,7 @@ import { useState } from "react";
  * @returns [storedValue, setValue]
  * Source: https://usehooks.com/useLocalStorage/
  */
-export default function useLocalStorage(key, initialValue) {
+export default function useLocalStorage(key, initialValue: any) {
```
MatoPlus: We can type this, this could help you catch some bugs as well:
```suggestion
export default function useLocalStorage<Type>(
    key: string,
    initialValue: Type,
): [Type, (value: Type) => void] {
```

utils/validation/fields.ts
MatoPlus: We can probably delete this file?

MatoPlus (changes requested):
Made a couple more suggestions, can you also type useLocalStorage while you are at it? I think it should just be

```typescript
export default function useLocalStorage<Type>(
    key: string,
    initialValue: Type,
): [Type, (value: Type) => void] {
```

as mentioned below. This should help you catch bugs in this huge diff 🤞

pages/parent/signup.tsx
gregmaxin: Thanks :)

pages/volunteer/signup.tsx
gregmaxin: Sure :)

utils/validation/fields.ts
gregmaxin: Yes, I'll delete this file

bekahma (commented):
looks good design wise! the error checking makes sense but is it possible to make it so it doesn't show by default? for example, only show when the user clicks into a field and clicks out again or when they try to go next before meeting the requirements. having them all show at once might be distracting/alarming

Screen Shot 2021-10-07 at 9 34 05 PM

@MatoPlus i'm not sure if this is already planned for another ticket but just thought i would mention it here

MatoPlus:
> @MatoPlus i'm not sure if this is already planned for another ticket but just thought i would mention it here
 
I agree with how the flow seems distracting. I'll make a ticket if we do not resolve this in this ticket.

gregmaxin:
> in terms of text field validations like postal code and phone number, everything looks good to me! @GodGreg i'm curious if there's a reason why a volunteer's name is not required? i know it's not explicitly stated in the designs but wondering what your thoughts are for that :) @MatoPlus @vedantdsn any thoughts?
> 
> Screen Shot 2021-10-05 at 8 38 44 PM

Currently it is required. Maybe you are a commit behind? 

image

gregmaxin:
> looks good design wise! the error checking makes sense but is it possible to make it so it doesn't show by default? for example, only show when the user clicks into a field and clicks out again or when they try to go next before meeting the requirements. having them all show at once might be distracting/alarming
> 
> Screen Shot 2021-10-07 at 9 34 05 PM
> 
> @MatoPlus i'm not sure if this is already planned for another ticket but just thought i would mention it here

I changed the field components to now contain a `interactedWith` state. This is what all the major form libraries do behind the scenes

gregmaxin:
> As part of this ticket, could you also make sure that the `/signup` page fetches serverside session prop for the wrapper to be handled as well? Looks like I forgot to patch that in.
> 
> image ^ always "signin" button.
> 
> In fact, we actually want to redirect people to the login page if they do not have a session.

```
    async function checkUserSession() {
        const request = {
            method: "GET",
            headers: { "Content-Type": "application/json" },
        };
        const response = await fetch("/api/user/me/", request).then(
            //To get the body do response.json()
            (response) => response,
        );
        if (!response.ok) {
            router.push("/login").then(() => window.scrollTo(0, 0));
        }
    }
    ```

Automatically redirects to /login if the user is on the signup page and not logged in

File: pages/signup.tsx [hunk 16]
```diff
@@ -21,6 +21,22 @@ export default function Signupform(): JSX.Element {
...
+            //To get the body do response.json()
+            (response) => response,
+        );
+        if (!response.ok) {
+            router.push("/login").then(() => window.scrollTo(0, 0));
+        }
+    }
+
+    checkUserSession();
+
```
MatoPlus: Close but not quite.... We want to use getServerProps to handle this on prerender, and it'll be much safer as well (like the class page for example).
```suggestion
```

We want to add the following function to the page:
```typescript
/**
 * getServerSideProps gets the session before this page is rendered
 */
export const getServerSideProps: GetServerSideProps = async (
    context: GetSessionOptions,
) => {
    // obtain the next auth session
    const session = await getSession(context);

    if (!session) {
        return {
            redirect: {
                destination: "/login",
                permanent: false,
            },
        };
    }

    return {
        props: { session },
    };
};
```

We also want to add the session into the wrapper component

```typescript
type SignupformProps = {
    session: Record<string, unknown>;
};

/**
 * This is the page that a user will use to either login or register
 * to the SDC platform as a parent of volunteer
 */
export default function Signupform({ session }: SignupformProps): JSX.Element {
    const router = useRouter();
    // hook to hold the url to redirect to
    const [url, setUrl] = useState("");

    const isUrlEmpty = () => {
        return url === "";
    };

    const isUrlPath = (path: string) => {
        return url === path;
    };

    return (
        <Wrapper session={session}>
...
```

MatoPlus (changes requested):
Made a suggestion that should fix build. Poke me if you want more details on this :)

pages/api/user/index.ts
```diff
@@ -18,6 +18,7 @@ export default async function handle(
     switch (req.method) {
         case "GET": {
             const users = await getUsers();
+            console.log(users);
```
MatoPlus:
```suggestion
```

models/Student.ts
```diff
@@ -20,11 +20,13 @@ export type CreateStudentInput = {
...
     allergies?: string;
     additionalInfo?: string;
-    emergFirstName: string;
-    emergLastName: string;
-    emergNumber: string;
-    emergRelationToStudent: string;
+    emergFirstName?: string;
+    emergLastName?: string;
+    emergNumber?: string;
+    emergRelationToStudent?: string;
```
MatoPlus: Looking at our schema, let's just make these fields required for now. So please make the corresponding validation in the front end!
```prisma
model Student {
  id                     Int            @id @default(autoincrement())
  parentId               Int            @map("parent_id")
  firstName              String         @map("first_name")
  lastName               String         @map("last_name")
  dateOfBirth            DateTime       @map("date_of_birth") @db.Timestamptz(6)
  addressLine1           String         @map("address_line1")
  addressLine2           String?        @map("address_line2")
  postalCode             String?        @map("postal_code") @db.VarChar(10)
  cityName               String?        @map("city_name")
  province               province?
  school                 String?
  grade                  Int?
  difficulties           difficulties[]
  otherDifficulties      String?        @map("other_difficulties")
  therapy                therapy[]
  otherTherapy           String?        @map("other_therapy")
  specialEducation       Boolean?       @default(false) @map("special_education")
  guardianExpectations   String?        @map("guardian_expectations")
  medication             String?
  allergies              String?
  additionalInfo         String?        @map("additional_info")
  emergFirstName         String         @map("emerg_first_name")
  emergLastName          String         @map("emerg_last_name")
  emergNumber            String         @map("emerg_number") @db.VarChar(50)
  emergRelationToStudent String         @map("emerg_relation_to_student")
  createdAt              DateTime       @default(now()) @map("created_at") @db.Timestamptz(6)
  updatedAt              DateTime?      @map("updated_at") @db.Timestamptz(6)
  parent                 Parent         @relation("parentsTostudents", fields: [parentId], references: [id])
  parentRegs             ParentReg[]

  @@map("students")
}
```

```suggestion
    emergFirstName: string;
    emergLastName: string;
    emergNumber: string;
    emergRelationToStudent: string;
```

pages/api/user/me.ts
```diff
@@ -17,8 +17,17 @@ export default async function handle(
     switch (req.method) {
         case "GET": {
             // This should be a me query
-            const user = await getUserFromEmail(session.user.email);
+            if (!session) {
+                ResponseUtil.returnBadRequest(res, "Invalid Session");
+            }
+            const user = await getUserFromEmail(session?.user.email);
+            console.log(user);
```
MatoPlus:
```suggestion
```

pages/api/user/me.ts
```diff
@@ -17,8 +17,17 @@ export default async function handle(
     switch (req.method) {
         case "GET": {
             // This should be a me query
-            const user = await getUserFromEmail(session.user.email);
+            if (!session) {
+                ResponseUtil.returnBadRequest(res, "Invalid Session");
```
MatoPlus, line 21: For this line in my branch I actually did something like:
```typescript
                ResponseUtil.returnOK(res, null);
```
Since returning no users is technically an ok response. It's not the same as if there is a session and we didn't get anything from our database. Let me know you thoughts!

MatoPlus (changes requested):
Made a couple more suggestions! This ticket is getting a bit bigger now! The document page has been merged into staging. If you do a rebase on staging, and integrate document uploads.

Feel free to pair with me or @matthewgeng to get this integrated! The idea is that we would want to redirect to `/document-upload?type=criminal-check` or `/document-upload?type=proof-of-income` which would upload the file to the corresponding path in our s3 bucket. After the pdf is submitted, it's currently setup such that the user can click the close button and it routes them back to where they came from (the signup page).

To keep track of which page we came form, we could even use query params like `/parent/signup?page=3` for example.

If this seems too much to take on, feel free to just skip the document upload integration, merge this in, and I can handle the rest 💪

pages/volunteer/signup.tsx
```diff
@@ -183,36 +234,30 @@ export default function VolunteerInfo({
...
         if (updatedUser) {
-            clearLocalStorage();
+            setSuccessfulAccountCreation("success");
+            localStorage.clear();
             return updatedUser;
         } else {
+            setSuccessfulAccountCreation("failure");
             return null;
         }
     }
```
MatoPlus: You might also want to invalidate the me query to let SWR know it's time to refetch the current user. Since our me query SWR call is `const { data, error } = useSWR("/api/user/me", fetcher);`, you would want to do

```typescript
    async function updateUserAndClearForm() {
        const updatedUser = await updateUser();
        if (updatedUser) {
            setSuccessfulAccountCreation("success");
            localStorage.clear();
            // New line to invalidate me query
            mutate("/api/user/me");
            return updatedUser;
        } else {
            setSuccessfulAccountCreation("failure");
            return null;
        }
    }
```


This is especially important once you rebase your branch, (there's me queries everywhere now lol)

pages/parent/signup.tsx
```diff
@@ -420,65 +519,33 @@ export default function ParticipantInfo({
...
-        setHeardFromFriendsAndFam(false);
-        setHeardFromFlyers(false);
-        setHeardFromEmail(false);
-        setHeardFromSocialMedia(false);
-        setHeardFromOther(false);
-        setHeardFromOptions([]);
-    };
-
+    //Form is finished
     async function updateUserAndClearForm() {
```
MatoPlus, line 529: idem, we want to add `mutate("/api/user/me")` after we updated the user.

MatoPlus (changes requested):

pages/volunteer/signup.tsx
gregmaxin: Sounds good

gregmaxin:
PR Changes

- [ ] Emergency section on parent form is all required
- [ ] Use the "me" queries
- [ ] Use `mutate` to clear the user causing a refresh after a successful account creation
- [ ] Fixed bug regarding a required checkbox on the learning disability page
- [ ] Navigating away 
- [ ] `const [pageNum, setPageNum] = useLocalStorage("VolunteerPage", 0);` to return to the previous page you were on. For example - navigating away for document upload

src/components/parent-form/ParentCreatedPage.tsx
```diff
@@ -93,16 +96,21 @@ export const ParentCreatedPage: React.FC<ParentCreatedPageProps> = ({
                         </Stack>
                     </Box>
                 </Center>
+            ) : successful === "pending" ? (
+                <Spinner />
```
MatoPlus: Let's use our Loading component instead!
```suggestion
                <Loading />
```

src/components/parent-form/ParticipantInfoPage.tsx
```diff
@@ -7,36 +7,49 @@ import {
...
+import moment from "moment";
+import colourTheme from "@styles/colours";
+import { PostalCodeField } from "@components/formFields/PostalCodeField";
+import { TextField } from "@components/formFields/TextField";
+import { ProvinceField } from "@components/formFields/ProvinceField";
+import { CheckBoxField } from "@components/formFields/CheckBoxField";
+import { DateField } from "@components/formFields/DateField";
 
 import { province } from "@models/User";
+import validator from "validator";
```
MatoPlus: There seems to be some unused imports. Please do an organize import on your files!

src/components/volunteer-form/VolunteerCreatedPage.tsx
```diff
@@ -90,17 +93,21 @@ export const VolunteerCreatedPage: React.FC<VolunteerCreatedPageProps> = ({
                         </Stack>
                     </Box>
                 </Center>
+            ) : successful === "pending" ? (
+                <Spinner />
```
MatoPlus: idem, let's use Loading here.
```suggestion
                <Loading />
```

src/components/parent-form/ParticipantInfoPage.tsx
```diff
@@ -51,110 +64,96 @@ export const ParticipantInfoPage: React.FC<ParticipantPageProps> = ({
...
                     value={props.dateOfBirth}
-                />
-            </FormControl>
-            <FormControl id="street-address-1">
-                <FormLabel>Street Address 1</FormLabel>
-                <Input
-                    placeholder="815 Hornby St."
-                    onChange={(e) => props.setAddress1(e.target.value)}
+                    setValue={props.setDateOfBirth}
+                ></DateField>
```
MatoPlus: I think the `DateField` comp can be self-closing
```suggestion
                />
```

src/components/volunteer-form/VolunteerDetailsPage.tsx
```diff
@@ -1,116 +1,108 @@
...
-                        placeholder="V6Z 2E6"
-                        onChange={(e) => props.setPostalCode(e.target.value)}
-                        value={props.postalCode}
-                    />
-                </FormControl>
+            <DateField
+                name={"Date Of Birth"}
+                value={props.dateOfBirth}
+                setValue={props.setDateOfBirth}
+            ></DateField>
```
MatoPlus: idem, self closing
```suggestion
            />
```

MatoPlus (approved):
LGTM! Great work 🔥 🔥 

There was a regression in the update user prisma functions. I tested everything with a rebase and things seem to be working great!

Just a few nitpicks from me and this is good to ship 🚢

MatoPlus:
Build is failing, remember to import Loading from `@components/Loading`

PR #86, opened 2021-10-05
Create Confirm Personal Info Page on Class Registration

ZhangCreations:
### Notion ticket link

[Create Confirm Personal Info Page on Class Registration](https://www.notion.so/uwblueprintexecs/Create-Confirm-Personal-Info-Page-on-Class-Registration-6023125d0b0a43339ec87298727ae488)

Implemented the page (/parent-enrollment) according to the Figma spec. Notably the _Relationship to Participant_ is missing under Parent/Guardian Information since that field doesn't exist in the db as far as I could find (but it is a question in registration, looking for some clarification on that)

### Steps to test

1. Create an account with a student
2. go to localhost:3000/parent-enrollment
3. Select a student
4. Confirm that all the information is displayed properly
5. Try it with various values under the _Forms of Therapy_ and _The participant has:_ sections (null values, all values etc.) since there is some funky calculation happening under the hood

- The aforementioned missing field
- The aforementioned fields with funky calculations

### Checklist

-   [x] I have run the linter

### Demo

https://user-images.githubusercontent.com/53587708/136139143-554ead7c-42b1-41eb-98e9-61ff0f0db2a8.mp4

### DIFF
--- a/pages/parent-enrollment.tsx
+++ b/pages/parent-enrollment.tsx
@@ -1,81 +1,75 @@
 import Wrapper from "@components/SDCWrapper";
-import { Flex, Button, Center, Text, VStack } from "@chakra-ui/react";
+import useSWR from "swr";
 import React, { useState } from "react";
-import { CloseButton } from "@components/CloseButton";
+import SelectChildForClass from "@components/SelectChildForClass";
+import { ClassEnrollmentConfirmation } from "@components/ClassEnrollmentConfirm";
+import { Text, Box } from "@chakra-ui/layout";
+import { GetServerSideProps } from "next";
+import { getSession, GetSessionOptions } from "next-auth/client";
 
-export default function ParentEnrollClass(): JSX.Element {
-    // Next button is disabled by default, activates when a child is selected
-    // Test data to be replaced with children associated with parent during integration
-    const children = ["Christina Ru", "Raewyn Tsai", "Stacy Kwok"];
-    const [selectedChild, setSelectedChild] = useState<string>("");
+type ParentEnrollClassProps = {
+    session: Record<string, unknown>;
+};
 
-    return (
-        <Wrapper>
-            <Flex justifyContent="flex-end">
-                {/* navigate to browse programs page instead of going back */}
-                <CloseButton href="/" />
-            </Flex>
-            <Center>
-                <Text align="center" mt="15px" fontWeight="700" fontSize="36px">
-                    Program Registration
-                </Text>
-            </Center>
-            <Center>
-                <Text pb="55px" align="center" mt="30px">
-                    Who would you like to register for{" "}
-                    <b>
-                        Building Bridges with Music - <br />
-                        Singing Monkeys (Ages 9 and under)?{" "}
-                    </b>
-                    (select one)
-                </Text>
-            </Center>
+export default function ParentEnrollClass(
+    props: ParentEnrollClassProps,
+): JSX.Element {
+    const [pageNum, setPagNum] = useState<number>(0);
+    const [selectedChild, setSelectedChild] = useState<number>(0);
 
-            <Center>
-                <VStack spacing={5}>
-                    {children.map((childName) => (
-                        <Button
-                            _focus={{ boxShadow: null }}
-                            _hover={{ border: "2px solid #0C53A0" }}
-                            backgroundColor={
-                                selectedChild === childName
-                                    ? "#E2E8F0"
-                                    : "white"
-                            }
-                            lineHeight="24px"
-                            fontSize="16px"
-                            fontWeight="normal"
-                            textColor="#0C53A0"
-                            borderRadius="6px"
-                            height="50px"
-                            width="340px"
-                            key={childName}
-                            onClick={() => setSelectedChild(childName)}
-                            border={
-                                selectedChild === childName
-                                    ? null
-                                    : "2px solid #E1E1E1"
-                            }
-                        >
-                            {childName}
-                        </Button>
-                    ))}
-                </VStack>
-            </Center>
-            <Center mt="45px" mb="200px">
-                <Button
-                    height="50px"
-                    width="340px"
-                    borderRadius="6px"
-                    background={selectedChild === "" ? "#737373" : "#0C53A0"}
-                    fontWeight="normal"
-                    textColor="#FFFFFF"
-                    fontSize="16px"
-                    isDisabled={selectedChild === ""}
-                >
-                    Next
-                </Button>
-            </Center>
-        </Wrapper>
+    const fetcher = (url) => fetch(url).then((res) => res.json());
+    const { data: apiResponse, error } = useSWR(
+        `/api/user/${props.session.id}`,
+        fetcher,
     );
+    if (error) {
+        return <Box>{"An error has occurred: " + error.toString()}</Box>;
+    }
+    if (apiResponse) {
+        try {
+            const userData = apiResponse.data;
+            const parentData = {
+                name: userData.firstName + " " + userData.lastName,
+                phone: userData.parent.phoneNumber,
+            };
+            const studentData = userData.parent.students;
+            const studentNames = userData.parent.students.map((s) => {
+                return `${s.firstName} ${s.lastName}`;
+            });
+            console.log(userData);
+            const pageElements = [
+                <SelectChildForClass
+                    children={studentNames}
+                    selectedChild={selectedChild}
+                    setSelectedChild={setSelectedChild}
+                    pageNum={pageNum}
+                    setPageNum={setPagNum}
+                />,
+                <ClassEnrollmentConfirmation
+                    studentData={studentData[selectedChild]}
+                    parentData={parentData}
+                    pageNum={pageNum}
+                    setPageNum={setPagNum}
+                />,
+            ];
+            return <Wrapper>{pageElements[pageNum]}</Wrapper>;
+        } catch {
+            console.log("There was an error");
+        }
+    }
+    return <Box></Box>;
 }
+
+/**
+ * getServerSideProps gets the session before this page is rendered
+ */
+export const getServerSideProps: GetServerSideProps = async (
+    context: GetSessionOptions,
+) => {
+    // obtain the next auth session
+    const session = await getSession(context);
+
+    return {
+        props: { session },
+    };
+};
--- a/pages/parent-enrollment.tsx
+++ b/pages/parent-enrollment.tsx
@@ -26,10 +26,6 @@ export default function ParentEnrollClass(
     const { user, isLoading, error } = useUser(props.session.id as string);
     const router = useRouter();
 
-    if (!props.session.role) {
-        router.push("/").then(() => window.scrollTo(0, 0)); // Redirect to home if user is not logged in
-    }
-
     if (error) {
         return <Box>{"An error has occurred: " + error.toString()}</Box>;
     }
@@ -79,6 +75,15 @@ export const getServerSideProps: GetServerSideProps = async (
 ) => {
     // obtain the next auth session
     const session = await getSession(context);
+
+    if (!session) {
+        return {
+            redirect: {
+                destination: "/login",
+                permanent: false,
+            },
+        };
+    }
 
     return {
         props: { session },
--- a/pages/parent-enrollment.tsx
+++ b/pages/parent-enrollment.tsx
@@ -5,23 +5,30 @@ import { ClassEnrollmentConfirmation } from "@components/ClassEnrollmentConfirm"
 import { Box } from "@chakra-ui/layout";
 import { GetServerSideProps } from "next";
 import useUser from "@utils/useUser";
+import { useRouter } from "next/router";
 import { getSession, GetSessionOptions } from "next-auth/client";
 import { Loading } from "@components/Loading";
+import { Student } from "@prisma/client";
 
 type ParentEnrollClassProps = {
     session: Record<string, unknown>;
 };
 
 /**
- * This is the page that a direct a user to register a student for a class
+ * This is the page that directs a user to register a student for a class
  */
 
 export default function ParentEnrollClass(
     props: ParentEnrollClassProps,
 ): JSX.Element {
-    const [pageNum, setPagNum] = useState<number>(0);
+    const [pageNum, setPageNum] = useState<number>(0);
     const [selectedChild, setSelectedChild] = useState<number>(0);
     const { user, isLoading, error } = useUser(props.session.id as string);
+    const router = useRouter();
+
+    if (!props.session.role) {
+        router.push("/").then(() => window.scrollTo(0, 0)); // Redirect to home if user is not logged in
+    }
 
     if (error) {
         return <Box>{"An error has occurred: " + error.toString()}</Box>;
@@ -29,34 +36,34 @@ export default function ParentEnrollClass(
     if (isLoading) {
         return <Loading></Loading>;
     }
-    if (user) {
-        const parentData = {
-            name: user.firstName + " " + user.lastName,
-            phone: user.parent.phoneNumber,
-        };
-        const studentData = user.parent.students;
-        const studentNames = user.parent.students.map((s) => {
-            return `${s.firstName} ${s.lastName}`;
-        });
-        console.log(studentData.dateOfBirth);
-        const pageElements = [
-            <SelectChildForClass
-                children={studentNames}
-                selectedChild={selectedChild}
-                setSelectedChild={setSelectedChild}
-                pageNum={pageNum}
-                setPageNum={setPagNum}
-            />,
-            <ClassEnrollmentConfirmation
-                studentData={studentData[selectedChild]}
-                parentData={parentData}
-                pageNum={pageNum}
-                setPageNum={setPagNum}
-            />,
-        ];
-        return <Wrapper>{pageElements[pageNum]}</Wrapper>;
+    const parentData = {
+        name: user.firstName + " " + user.lastName,
+        phone: user.parent.phoneNumber,
+    };
+    const studentData = user.parent.students;
+    const studentNames = studentData.map((s) => {
+        return `${s.firstName} ${s.lastName}`;
+    });
+
+    if (studentData.length < 1) {
+        router.push("/").then(() => window.scrollTo(0, 0)); // Redirect to home if there are no children, this should be updated to a toast in a future sprint
     }
-    return <Box></Box>;
+    const pageElements = [
+        <SelectChildForClass
+            children={studentNames}
+            selectedChild={selectedChild}
+            setSelectedChild={setSelectedChild}
+            pageNum={pageNum}
+            setPageNum={setPageNum}
+        />,
+        <ClassEnrollmentConfirmation
+            studentData={studentData[selectedChild] as Student}
+            parentData={parentData}
+            pageNum={pageNum}
+            setPageNum={setPageNum}
+        />,
+    ];
+    return <Wrapper>{pageElements[pageNum]}</Wrapper>;
 }
 
 /**
--- a/pages/parent-enrollment.tsx
+++ b/pages/parent-enrollment.tsx
@@ -40,7 +40,12 @@ export default function ParentEnrollClass(
         name: user.firstName + " " + user.lastName,
         phone: user.parent.phoneNumber,
     };
-    const studentData = user.parent.students;
+    const studentData = user.parent.students.map((s) => {
+        const replaceDate = new Date(s.dateOfBirth);
+        s.dateOfBirth = replaceDate;
+        return s;
+    });
+
     const studentNames = studentData.map((s) => {
         return `${s.firstName} ${s.lastName}`;
     });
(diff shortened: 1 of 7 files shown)


pages/parent-enrollment.tsx
```diff
@@ -1,81 +1,75 @@
...
-                <Text pb="55px" align="center" mt="30px">
-                    Who would you like to register for{" "}
-                    <b>
-                        Building Bridges with Music - <br />
-                        Singing Monkeys (Ages 9 and under)?{" "}
-                    </b>
-                    (select one)
-                </Text>
-            </Center>
+export default function ParentEnrollClass(
```
MatoPlus, line 21: Can we add a small header doc comment here? Look at the other pages for example!

I was thinking maybe something like:
```suggestion
/**
 * This is the page handle parent enrollment into classes, through the entire registration flow
 */
export default function ParentEnrollClass(
```

pages/parent-enrollment.tsx
```diff
@@ -1,81 +1,75 @@
...
-                >
-                    Next
-                </Button>
-            </Center>
-        </Wrapper>
+    const fetcher = (url) => fetch(url).then((res) => res.json());
+    const { data: apiResponse, error } = useSWR(
+        `/api/user/${props.session.id}`,
+        fetcher,
     );
```
MatoPlus: It would be awesome if you can make this an encapsulated useSWR hook, if you do `git fetch && git rebase origin/staging`, resolve the merge conflicts, you'll have some examples in the `utils` folder. 

In this case, we might want something like

```typescript
export default function useUser(id: string) {
    const { data, error } = useSWR(
        `/api/user/${id}`,
        fetcher,
    );
    return {
        user: data ? data.data : null,
        isLoading: !error && !data,
        error: error,
    };
}
```

pages/parent-enrollment.tsx
```diff
@@ -1,81 +1,75 @@
...
+                    studentData={studentData[selectedChild]}
+                    parentData={parentData}
+                    pageNum={pageNum}
+                    setPageNum={setPagNum}
+                />,
+            ];
+            return <Wrapper>{pageElements[pageNum]}</Wrapper>;
+        } catch {
+            console.log("There was an error");
+        }
```
MatoPlus: I think if we use the hook I mentioned above, we can have three cases, 1. if error, render what box with error, 2. if loading, maybe use the new Loading component in `origin/staging`. 3. simply just render what you have in the expected case.

So we would not need the try catch here, we want to try and avoid these kind of things, doesn't read well.

src/components/ClassEnrollmentConfirm.tsx
```diff
@@ -0,0 +1,406 @@
...
+    Text,
+    Stack,
+    Flex,
+} from "@chakra-ui/react";
+import { BackButton } from "@components/BackButton";
+import { CloseButton } from "@components/CloseButton";
+
+type ClassEnrollmentConfirmationProps = {
+    studentData: any; // Not actually any but I don't want to type the entire student object, let me know if I have to do this or if it's done somewhere else that can be used here
+    parentData: any; // ^^^ Same idea
```
MatoPlus: Smart! I think also just use the types that the prisma client generated for us, for student data, you could just use `Student from @prisma/client` and you can use User type from parentData (then for this above, you can just pass in the raw userData into this prop). Pretty neat huh.

src/components/ClassEnrollmentConfirm.tsx
```diff
@@ -0,0 +1,406 @@
...
+            <VStack spacing="45px" alignItems="flex-start">
+                <HStack
+                    alignSelf="flex-start"
+                    mt="15px"
+                    fontWeight="700"
+                    fontSize="36px"
+                >
+                    <Heading>
+                        Confirm Personal Information{" "}
+                        <Heading color="#0C53A0" as="span">
```
MatoPlus: Please import `colourTheme` and use `colourTheme.colors.Blue` instead, one source of truth.

src/components/ClassEnrollmentConfirm.tsx
```diff
@@ -0,0 +1,406 @@
...
+                    alignItems="flex-start"
+                    spacing="32px"
+                    marginBottom="80px"
+                >
+                    <Text fontWeight={700} fontSize="24px" marginBottom="-16px">
+                        General Participant Information
+                    </Text>
+                    <Box>
+                        <Text
+                            color="#737373"
```
MatoPlus: Please add a Gray to our colourTheme and use it here.

src/components/ClassEnrollmentConfirm.tsx
```diff
@@ -0,0 +1,406 @@
...
+                            Participant Name{" "}
+                        </Text>
+                        <Text>
+                            {props.studentData.firstName}{" "}
+                            {props.studentData.lastName}
+                        </Text>
+                    </Box>
+                    <Box>
+                        <Text
+                            color="#737373"
```
MatoPlus: idem, same with all the other instance

src/components/SelectChildForClass.tsx
```diff
@@ -0,0 +1,94 @@
...
+                        </Button>
+                    ))}
+                </VStack>
+            </Center>
+            <Center mt="45px" mb="200px">
+                <Button
+                    height="50px"
+                    width="340px"
+                    borderRadius="6px"
+                    background="#0C53A0"
```
MatoPlus: colourTheme
```suggestion
                    background={colourTheme.colors.Blue}
```

src/components/SelectChildForClass.tsx
```diff
@@ -0,0 +1,94 @@
...
+                            _hover={{ border: "2px solid #0C53A0" }}
+                            backgroundColor={
+                                props.selectedChild === index
+                                    ? "#E2E8F0"
+                                    : "white"
+                            }
+                            lineHeight="24px"
+                            fontSize="16px"
+                            fontWeight="normal"
+                            textColor="#0C53A0"
```
MatoPlus: colourTheme
```suggestion
                            textColor={colourTheme.colors.Blue}
```

src/components/SelectChildForClass.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+            <Center>
+                <VStack spacing={5}>
+                    {props.children.map((childName, index) => (
+                        <Button
+                            _focus={{ boxShadow: null }}
+                            _hover={{ border: "2px solid #0C53A0" }}
+                            backgroundColor={
+                                props.selectedChild === index
+                                    ? "#E2E8F0"
```
MatoPlus: I think we have a colourTheme for this in origin/staging, otherwise add it please

src/components/ClassEnrollmentConfirm.tsx
```diff
@@ -0,0 +1,406 @@
...
+                        >
+                            {" "}
+                            Emergency Contact Cell Number{" "}
+                        </Text>
+                        <Text>
+                            {props.studentData.emergNumber.slice(0, 3) +
+                                "-" +
+                                props.studentData.emergNumber.slice(3, 6) +
+                                "-" +
+                                props.studentData.emergNumber.slice(6, 10)}
```
MatoPlus: I think we can make this a util.

src/components/ClassEnrollmentConfirm.tsx
```diff
@@ -0,0 +1,406 @@
...
+                        </Box>
+                    </Stack>
+                    <Box>
+                        <Text
+                            color="#737373"
+                            fontSize="14px"
+                            marginBottom="6px"
+                        >
+                            {" "}
+                            The participant has{" "}
```
MatoPlus: Missing a colon but it does look a bit off.
@bekahma @vedantdsn 
Figma:
image
Thoughts if we it's fine to remove the colon here?

MatoPlus (changes requested):
💪  Great work Kevin! Just a few nit picks from me.

Yea It's kinda funny how we are missing the parent relationship field, I don't think our forms our actually accounting for full parent info at all ATM, which is kinda funny, kinda related problem in validation forms as well (forms to save parent name). We might want to add a DB column and add a new parent form page in `parent/signup` (another ticket).

@bekahma @vedantdsn might wanna see what's up. In the parent signup form, they never actually enter own info (including parent name, just emergency contacts) If we can't come up with a quick design tweak, maybe we can just roll with no parent relation for now.

I'll come back to this later today.

src/components/SelectChildForClass.tsx
```diff
@@ -0,0 +1,95 @@
...
+                    (select one)
+                </Text>
+            </Center>
+
+            <Center>
+                <VStack spacing={5}>
+                    {props.children.map((childName, index) => (
+                        <Button
+                            _focus={{ boxShadow: null }}
+                            _hover={{ border: "2px solid #0C53A0" }}
```
MatoPlus: I think you get can away with:
```suggestion
                            _hover={{ border: `2px solid ${colourTheme.colors.Blue}` }}
```

src/components/SelectChildForClass.tsx
```diff
@@ -0,0 +1,95 @@
...
+    setPageNum: React.Dispatch<SetStateAction<number>>;
+};
+
+export default function SelectChildForClass(
+    props: SelectChildForClassProps,
+): JSX.Element {
+    // Next button is disabled by default, activates when a child is selected
+    // Test data to be replaced with children associated with parent during integration
+    // const children = ["Christina Ru", "Raewyn Tsai", "Stacy Kwok"];
+    // const [selectedChild, setSelectedChild] = useState<string>("");
```
MatoPlus: Please remove this.
```suggestion
```

src/components/ClassEnrollmentConfirm.tsx
```diff
@@ -0,0 +1,403 @@
...
+    Flex,
+} from "@chakra-ui/react";
+import { BackButton } from "@components/BackButton";
+import { CloseButton } from "@components/CloseButton";
+import { Student, Parent } from "@prisma/client";
+import colourTheme from "@styles/colours";
+import parsePhoneNumber from "@utils/parsePhoneNumber";
+
+type ClassEnrollmentConfirmationProps = {
+    studentData: Student; // Not actually any but I don't want to type the entire student object, let me know if I have to do this or if it's done somewhere else that can be used here
```
MatoPlus:
```suggestion
    studentData: Student;
```

pages/parent-enrollment.tsx
```diff
@@ -1,82 +1,74 @@
...
+        return <Box>{"An error has occurred: " + error.toString()}</Box>;
+    }
+    if (isLoading) {
+        return <Loading></Loading>;
+    }
+    if (user) {
+        const parentData = {
+            name: user.firstName + " " + user.lastName,
+            phone: user.parent.phoneNumber,
+        };
```
MatoPlus: I think you don't need this. Just pass in user as Type User. And make the according changes in the component that uses this data.
```suggestion
```

pages/parent-enrollment.tsx
```diff
@@ -1,82 +1,74 @@
...
+            <SelectChildForClass
+                children={studentNames}
+                selectedChild={selectedChild}
+                setSelectedChild={setSelectedChild}
+                pageNum={pageNum}
+                setPageNum={setPagNum}
+            />,
+            <ClassEnrollmentConfirmation
+                studentData={studentData[selectedChild]}
+                parentData={parentData}
```
MatoPlus:
```suggestion
                parentData={user}
```

src/components/ClassEnrollmentConfirm.tsx
```diff
@@ -0,0 +1,403 @@
...
+    parentData: any; // ^^^ Same idea
```
MatoPlus: Use the user type from `@prisma/client`
```suggestion
    parentData: User
```

pages/parent-enrollment.tsx
```diff
@@ -1,82 +1,74 @@
...
+                pageNum={pageNum}
+                setPageNum={setPagNum}
+            />,
+        ];
+        return <Wrapper>{pageElements[pageNum]}</Wrapper>;
+    }
+    return <Box></Box>;
```
MatoPlus: I think this is stale, at this point, if not error or loading, then user has to exist, this code is never used.
```suggestion
```

pages/parent-enrollment.tsx
```diff
@@ -1,82 +1,74 @@
...
+    const [selectedChild, setSelectedChild] = useState<number>(0);
+    const { user, isLoading, error } = useUser(props.session.id as string);
+
+    if (error) {
+        return <Box>{"An error has occurred: " + error.toString()}</Box>;
+    }
+    if (isLoading) {
+        return <Loading></Loading>;
+    }
+    if (user) {
```
MatoPlus: If statement is redundant.
```suggestion
```

MatoPlus (changes requested):
Looks good to me, just a couple more cleanup items before we can get this in, I think there were also a few unresolved items from before.

src/components/ClassEnrollmentConfirm.tsx
bekahma: "The participant has" sounds awkward to me in general. I think it's good to remove the colon for now and @vedantdsn and I can see if we can come up with another way to phrase that

pages/parent-enrollment.tsx
ZhangCreations: I would do that, but the type User doesn't have the nested student and parent objects so the object user isn't of type User according to TS

pages/parent-enrollment.tsx
```diff
@@ -1,82 +1,74 @@
...
-                    Who would you like to register for{" "}
-                    <b>
-                        Building Bridges with Music - <br />
-                        Singing Monkeys (Ages 9 and under)?{" "}
-                    </b>
-                    (select one)
-                </Text>
-            </Center>
+/**
+ * This is the page that a direct a user to register a student for a class
```
LenaNguyen:
```suggestion
 * This is the page that directs a user to register a student for a class
```

pages/parent-enrollment.tsx
```diff
@@ -1,82 +1,74 @@
...
-                >
-                    Next
-                </Button>
-            </Center>
-        </Wrapper>
-    );
+export default function ParentEnrollClass(
+    props: ParentEnrollClassProps,
+): JSX.Element {
+    const [pageNum, setPagNum] = useState<number>(0);
```
LenaNguyen: nit: `setPageNum` (missing the e)

pages/parent-enrollment.tsx
```diff
@@ -1,82 +1,74 @@
...
+    if (user) {
+        const parentData = {
+            name: user.firstName + " " + user.lastName,
+            phone: user.parent.phoneNumber,
+        };
+        const studentData = user.parent.students;
+        const studentNames = user.parent.students.map((s) => {
+            return `${s.firstName} ${s.lastName}`;
+        });
+        console.log(studentData.dateOfBirth);
```
LenaNguyen:
```suggestion

```

pages/parent-enrollment.tsx
```diff
@@ -1,82 +1,74 @@
...
+    if (isLoading) {
+        return <Loading></Loading>;
+    }
+    if (user) {
+        const parentData = {
+            name: user.firstName + " " + user.lastName,
+            phone: user.parent.phoneNumber,
+        };
+        const studentData = user.parent.students;
+        const studentNames = user.parent.students.map((s) => {
```
LenaNguyen:
```suggestion
        const studentNames = studentData.map((s) => {
```

src/components/ClassEnrollmentConfirm.tsx
```diff
@@ -0,0 +1,403 @@
...
+    pageNum: number;
+    setPageNum: React.Dispatch<SetStateAction<number>>;
+};
+
+const therapyMapping = {
+    PHYSIO: "Physiotherapy",
+    SPEECH_LANG: "Speech and language therapy",
+    OCCUPATIONAL: "Occupational therapy",
+    COUNSELLING: "Counseling",
```
LenaNguyen:
```suggestion
    COUNSELLING: "Counselling",
```

src/components/ClassEnrollmentConfirm.tsx
```diff
@@ -0,0 +1,403 @@
...
+                        General Participant Information
+                    </Text>
+                    <Box>
+                        <Text
+                            color={colourTheme.colors.Gray}
+                            fontSize="14px"
+                            marginBottom="6px"
+                        >
+                            {" "}
+                            Participant Name{" "}
```
LenaNguyen: Are these spaces just for adding margin-left and right? it seems pretty weird not to just do `marginLeft` and `marginRight` in the Text component.

Btw @MatoPlus looks like the codebase could benefit from creating a reusable component called `FormLabel` or something with the color, fontsize,  and margins preset (and overridable) so we don't have to do so much copying and pasting

LenaNguyen (changes requested):

LenaNguyen (approved):
Looks great! There's still a couple of Rickson's comments to resolve and designer review needed.  Nice job incorporating all the changes so quickly 🙌

src/components/ClassEnrollmentConfirm.tsx
MatoPlus: > Are these spaces just for adding margin-left and right? it seems pretty weird not to just do `marginLeft` and `marginRight` in the Text component.
> 
> Btw @MatoPlus looks like the codebase could benefit from creating a reusable component called `FormLabel` or something with the color, fontsize, and margins preset (and overridable) so we don't have to do so much copying and pasting

I'll make a ticket!

pages/parent-enrollment.tsx
```diff
@@ -1,82 +1,86 @@
...
+    props: ParentEnrollClassProps,
+): JSX.Element {
+    const [pageNum, setPageNum] = useState<number>(0);
+    const [selectedChild, setSelectedChild] = useState<number>(0);
+    const { user, isLoading, error } = useUser(props.session.id as string);
+    const router = useRouter();
+
+    if (!props.session.role) {
+        router.push("/").then(() => window.scrollTo(0, 0)); // Redirect to home if user is not logged in
+    }
```
MatoPlus: Sorry, my bad. I don't actually want this because it's not safe. Next.js could render this page server side and router would not be available as it can only be used client side. We want to make use of the getServerSideProps instead, see changes in the other comment.

```suggestion
```

pages/parent-enrollment.tsx
```diff
@@ -1,82 +1,86 @@
...
+
+/**
+ * getServerSideProps gets the session before this page is rendered
+ */
+export const getServerSideProps: GetServerSideProps = async (
+    context: GetSessionOptions,
+) => {
+    // obtain the next auth session
+    const session = await getSession(context);
+
```
MatoPlus, line 78: This redirects the user if they are not logged in, which can be done server side fast and safely. We can even consider only parent should be access this page as well:

```suggestion
    // obtain the next auth session
    const session = await getSession(context);
    
    if (!session || session.role !== roles.PARENT) {
        return {
            redirect: {
                destination: "/login",
                permanent: false,
            },
        };
    }
```

src/components/ClassEnrollmentConfirm.tsx
```diff
@@ -0,0 +1,390 @@
...
+    SPEECH_LANG: "Speech and language therapy",
+    OCCUPATIONAL: "Occupational therapy",
+    COUNSELLING: "Counselling",
+    ART: "Art therapy",
+};
+
+export const ClassEnrollmentConfirmation = (
+    props: ClassEnrollmentConfirmationProps,
+): JSX.Element => {
+    console.log(typeof props.studentData.dateOfBirth);
```
MatoPlus: I'm guessing not needed anymore. PROTIP: you can hover over variables to get their types much faster, unless they are an any/unknown type.
```suggestion
```

utils/parsePhoneNumber.ts
```diff
@@ -0,0 +1,8 @@
+/*
+ * @param s a string is in the format ##########
+ * @return phoneNUmber a string in the format ###-###-####
+ */
+
+export default function parsePhoneNumber(s: string) {
```
MatoPlus: add a return type here.

```suggestion
export default function parsePhoneNumber(s: string): string {
```

MatoPlus (changes requested):
I think this is the last few issues I have. I'll give stamp after changes are made. Don't forget about the unresolved comments I made in my first review (using `colourTheme`)!

pages/parent-enrollment.tsx
ZhangCreations: Oh thanks, I had no idea my implementation was unsafe but it's great to learn :)

src/components/ClassEnrollmentConfirm.tsx
ZhangCreations: This variable was being super weird, at runtime it was a string but the compiler thinks its a date so I had to double check what it was at each state.

ZhangCreations:
Addressed the PR comments and also fixed a bug with date of birth.

src/components/ClassEnrollmentConfirm.tsx
MatoPlus: ah that's pretty wack, glad you got it though

MatoPlus (approved):
LGTM, Ship it 🚢

PR #90, opened 2021-10-08
feat(class): user/student class eligibility

MatoPlus:
### Notion ticket link

[If user is logged in determine they are eligible for class](https://www.notion.so/uwblueprintexecs/If-user-is-logged-in-determine-they-are-eligible-for-class-6b3003074ac24769a10e8803f5a215ad)

https://user-images.githubusercontent.com/44826218/136488879-9db6688d-6a6b-41b5-95bb-5b49127be4ef.mp4

Full classes section:
image

Volunteer view:
image

Collapse:
https://user-images.githubusercontent.com/44826218/137058021-b4e2c7de-c260-4e5b-82de-f8f068b6f5f0.mp4

- If user (parent) is logged in determine their children are eligible for class. If they are not, the class badge should be grayed out and clicking on the class will prompt user that they are not eligible. If at least one of the children are eligible, the badge should be blue
- Divide class list into two parts (full and non-full) via filtering on availability
- Add volunteer specific view

### Steps to test

1. Test the flow in the implementation description as parent with many cases
    a. 0 child
    b. 1 child eligible for 1 class
    c. multiple children eligible
    c. multiple children not eligible
2. Test as non-logged in user
3. Test as logged-in volunteer

- Design: Look and feel of class ineligibility flow
- Design: Flow difference between parent/non-parent users
- Dev: New unstated-next usage for state management
- Dev: Logic of eligibility
- Dev: Code style
- Dev: DB changes

### Checklist

-   [X] I have run the linter

### DIFF
--- a/models/Class.ts
+++ b/models/Class.ts
@@ -22,7 +22,9 @@ export type ClassCardInfo = {
     image?: string;
     name?: string;
     description: string;
-    ageGroup?: string;
+    // ageGroup?: string;
+    borderAge?: number;
+    isAgeMinimal: boolean;
     spaceAvailable: number;
     spaceTotal: number;
     volunteerSpaceAvailable: number;
--- a/models/Class.ts
+++ b/models/Class.ts
@@ -5,7 +5,8 @@ import { weekday } from "@prisma/client";
  */
 export type ClassInput = {
     name?: string;
-    ageGroup?: string;
+    borderAge?: number;
+    isAgeMinimal: boolean;
     programId: number;
     spaceTotal: number;
     volunteerSpaceTotal: number;
@@ -22,7 +23,6 @@ export type ClassCardInfo = {
     image?: string;
     name?: string;
     description: string;
-    // ageGroup?: string;
     borderAge?: number;
     isAgeMinimal: boolean;
     spaceAvailable: number;
--- a/pages/api/user/me.ts
+++ b/pages/api/user/me.ts
@@ -16,6 +16,12 @@ export default async function handle(
 
     switch (req.method) {
         case "GET": {
+            // If no session, return no result
+            if (!session) {
+                ResponseUtil.returnOK(res, null);
+                return;
+            }
+
             // This should be a me query
             const user = await getUserFromEmail(session.user.email);
             ResponseUtil.returnOK(res, user);
--- a/pages/program-details/[pid].tsx
+++ b/pages/program-details/[pid].tsx
@@ -1,6 +1,6 @@
 import { useRouter } from "next/router";
 import React from "react";
-import { , Box from "@chakra-ui/react";
+import { Box } from "@chakra-ui/react";
 import { getSession } from "next-auth/client";
 import { ProgramInfo } from "@components/ProgramInfo";
 import useSWR from "swr";
@@ -10,6 +10,8 @@ import { GetServerSideProps } from "next";
 import { serverSideTranslations } from "next-i18next/serverSideTranslations";
 import { locale } from "@prisma/client";
 import { Loading } from "@components/Loading";
+import Participants from "@utils/containers/Participants";
+import useMe from "@utils/useMe";
 
 type ProgramDetailsProps = {
     session: Record<string, unknown>;
@@ -20,6 +22,8 @@ export const ProgramDetails: React.FC<ProgramDetailsProps> = ({
 }: ProgramDetailsProps) => {
     const router = useRouter();
     const { pid } = router.query;
+
+    const { me, isLoading: isMeLoading } = useMe();
 
     const { data: classListResponse, error: classListError } = useSWR(
         ["/api/class", pid, router.locale],
@@ -32,7 +36,7 @@ export const ProgramDetails: React.FC<ProgramDetailsProps> = ({
     );
     const isProgramInfoLoading = !programInfoResponse && !programInfoError;
 
-    if (isClassListLoading || isProgramInfoLoading) {
+    if (isClassListLoading || isProgramInfoLoading || isMeLoading) {
         return <Loading />;
     } else if (classListError || programInfoError) {
         return <Box>An Error has occured</Box>;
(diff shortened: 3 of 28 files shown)


bekahma (commented):
the flow looks good for both eligible and not eligible classes :) however, i noticed the badge colour (blue vs. grey) is not super obvious and it would be nice to have the photo be greyed out as well. maybe the `filter` property for `grayscale(100%)` (or something similar) could work here?

MatoPlus:
> the flow looks good for both eligible and not eligible classes :) however, i noticed the badge colour (blue vs. grey) is not super obvious and it would be nice to have the photo be greyed out as well. maybe the `filter` property for `grayscale(100%)` (or something similar) could work here?

Resolved (i hope)! @bekahma Thanks for catching that so quickly. Appreciated the css tip 👀

MatoPlus:
@bekahma @vedantdsn Sorry, can I get another stamp? I added two new things that I would like you guys to look at 👀 
- Full class section
- Volunteer view 
- Full class Collapse

pages/api/user/me.ts
```diff
@@ -16,6 +16,12 @@ export default async function handle(
 
     switch (req.method) {
         case "GET": {
+            // If no session, return no result
+            if (!session) {
+                ResponseUtil.returnOK(res, null);
```
feng-eric, line 21: Should it be a status code 200?

prisma/schema.sql
```diff
@@ -63,7 +63,8 @@ CREATE TABLE programs (
 CREATE TABLE classes (
   id SERIAL PRIMARY KEY NOT NULL,
   name TEXT, 
-  age_group TEXT, -- classes are always categorized by age group
+  border_age INTEGER NOT NULL,
```
feng-eric, line 70: Remember to reset staging DB after you merge this in

src/components/ClassInfoCard.tsx
```diff
@@ -27,8 +29,10 @@ type ClassInfoProps = {
  */
 export const ClassInfoCard: React.FC<ClassInfoProps> = ({
     cardInfo,
+    isLegible,
```
feng-eric: Can you explain what `isLegible` is for? When is it false (not readable). A bit confused on what readability has to do with the context

MatoPlus: yea should be `eligible` mb

pages/api/user/me.ts
MatoPlus: hmm, my thought was no session is not the same case as there is a session but we could not find a corresponding user. I would think no session is a valid 200 return.

models/Class.ts
```diff
@@ -22,7 +23,8 @@ export type ClassCardInfo = {
     image?: string;
     name?: string;
     description: string;
-    ageGroup?: string;
+    borderAge?: number;
+    isAgeMinimal: boolean;
```
feng-eric, line 27: Could you provide some documentation somewhere about how `borderAge` and `isAgeMinimal` is used? I think maybe in `schema.sql` (or somewhere else if you have a better idea)?

feng-eric (approved):
LGTM, had one small comment

models/Class.ts
MatoPlus: Will do! I think schema.sql is fine 👍

bekahma (approved):

pages/program-details/[pid].tsx
```diff
@@ -1,16 +1,21 @@
...
+import { Loading } from "@components/Loading";
+import Participants from "@utils/containers/Participants";
 
-export const ProgramDetails: React.FC = () => {
-    const [session, loading] = useSession();
+const ProgramDetails: React.FC = () => {
     const router = useRouter();
     const { pid } = router.query;
 
+    const { me, isLoading } = useMe();
```
Comment (LenaNguyen) [inline pages/program-details/[pid].tsx]: Nit:
```suggestion
    const { me, isLoading: isMeLoading } = useMe();
```

pages/program-details/[pid].tsx
```diff
@@ -1,16 +1,21 @@
...
+    const [session, loading] = useSession();
```
Comment (LenaNguyen) [inline pages/program-details/[pid].tsx]: Nit: makes the variable names less ambiguous

```suggestion
    const [session, isSessionLoading] = useSession();
```

pages/program-details/[pid].tsx
```diff
@@ -21,33 +26,34 @@ export const ProgramDetails: React.FC = () => {
...
-        return (
-            <Text>
-                An error has occurred.{" "}
-                {classListError
-                    ? classListError.toString()
-                    : programInfoError.toString()}
-            </Text>
-        );
+
+    if (isLoading || loading) {
```
Comment (LenaNguyen) [inline pages/program-details/[pid].tsx]: If you go with above then
```suggestion
    if (isMeLoading || isSessionLoading) {
```

src/components/ProgramInfo.tsx
```diff
@@ -29,6 +41,23 @@ export const ProgramInfo: React.FC<ProgramDetailsProps> = ({
...
+    const { me } = useMe();
+    let fullClassInfo;
+    let availableClassInfo;
+    if (me && me.volunteer) {
+        fullClassInfo = classInfo.filter(
+            (info) => info.volunteerSpaceAvailable === 0,
+        );
+        availableClassInfo = classInfo.filter(
+            (info) => info.volunteerSpaceAvailable !== 0,
+        );
```
LenaNguyen, line 61: you should be able to do this in one pass of the array instead of two using `forEach` or just a `for` loop like:
```
let fullClassInfo = [];
let availableClassInfo = [];
classInfo.forEach((info) => (info.volunteerSpaceAvailable === 0 ? fullClassInfo : availableClassInfo).push(info));
```

same for in the else block

LenaNguyen (approved):
approving bc there's just a couple minor things 🚢

src/components/ProgramInfo.tsx
MatoPlus: that's mega brain 👀

PR #106, opened 2021-10-19
Finalized Parent Enrollment

ZhangCreations:
### Notion ticket link

[Finalized Parent Enrollment](https://www.notion.so/uwblueprintexecs/Finalize-Parent-Enrollment-269c8281ab4d4fa58036d3128b153f6f)

- Connected all the components of the parent class enrollment
- Created discount applied page according to the design on Figma
- Added API call on last page to actually create the registration record

### Steps to test

1. Log in, ensure that the user you log in as has a child that can register for one of the courses in your test data.
2. Try to register for a course through one of the program cards, if that doesn't work go to `parent/enrollment?classId={VALID_CLASS_ID}`
3. Go through all the pages
4. Try all the options for the proof on income page as follows.
5. On the user object, if `isLowIncome = false` and there is no proof of income, then it should lead you to the proof of income page
6. Manually make your user `isLowIncome = true`, this should bring you to the discount page
7. Put a random string into the `proofOfIncome` field of the user , this should skip the proof of income page
8. Continue to the next page, then check the Db to ensure that there was a new registration record created

- There are a few things that I don't love in my implementation, that I could not resolve myself, I will point out here to see if anyone else can figure them out.
    -  In `ProofOfIncomeWrapper.tsx` the logic for `isLowIncome = false` and there is no proof of income should be to skip this page, thus the nextPage() function is called, however this needs to be guarded by the page being on the correct page number or else the rendering will cause the flow to skip to the end. Thus the page number of the income page is hard-coded, and I am not sure of a way to work around it. (line 94 of the file)
    - Once the user uploads their income proof if applicable, they are redirected to the beginning of the flow, I am not sure if that is an acceptable flow, or how to reasonable mitigate it (It might be possible by saving a very specialized value in state and checking when rendering this flow but I am not sure it's worth it)
    - The last page seems to make the API call multiple times when the page is left idle, I am not sure why this happens and how to circumvent it, but since there is the unique guard on the DB, it functions as expected.

### Checklist

-   [x] I have run the linter

### Demo

https://user-images.githubusercontent.com/53587708/138224159-a9697b05-91bb-44a8-9d4a-db49b444a020.mp4


pages/parent/enrollment.tsx
```diff
@@ -27,6 +28,8 @@ export default function ParentEnrollClass({
     const [selectedChild, setSelectedChild] = useState<number>(0);
     const { user, isLoading, error } = useUser(session.id as string);
     const router = useRouter();
+    const { urlArg } = router.query;
+    const classId = urlArg ? Number(urlArg) : null;
```
MatoPlus: we're passing in `classId` into router.query right? not `urlArg`.

We also want to have a page param to jump to pages after uploading a document. We want to set it as the default page count when the page loads:

```typescript
const [pageNum, setPageNum] = useState<number>(page ? parseInt(page, 10) : 0);
```

Maybe if we don't have a classId, we can just do a `router.back()` instead.

```suggestion
    const { classId, page } = router.query;
  
```

src/components/registration-form/DiscountPage.tsx
```diff
@@ -0,0 +1,74 @@
...
+                    <Heading fontSize="22px" fontWeight="900">
+                        Has your income changed?
+                    </Heading>
+                    <Button
+                        variant="ghost"
+                        as="u"
+                        onClick={() => {
+                            setUpdateProofOfIncome(true);
+                        }}
+                        // borderRadius="6px"
```
MatoPlus:
```suggestion
```

src/components/registration-form/ProofOfIncomeWrapper.tsx
```diff
@@ -0,0 +1,96 @@
...
+                        UPLOADING_PROOF_OF_INCOME={UPLOADING_PROOF_OF_INCOME}
+                        PROOF_OF_INCOME_EXAMPLES={PROOF_OF_INCOME_EXAMPLES}
+                    />
+                </Stack>
+                <Box>
+                    <HStack spacing="24px">
+                        <FormButton
+                            onClick={() => {
+                                router
+                                    .push("/document-upload?type=income-proof")
```
MatoPlus: We want to use document upload page with redirect (you'll need to pass in classId here):

Let me know if you want me to sync with you on this! And I can quickly talk with you about how this works.

```suggestion
                                    .push("/document-upload?type=income-proof&redirect=/parent/enrollment?classId=${classInfo.id}%26page=${
                                    pageNum + 1
                                }`,")
```

src/components/registration-form/ProofOfIncomeWrapper.tsx
```diff
@@ -0,0 +1,96 @@
...
+                                    .then(() => window.scrollTo(0, 0));
+                            }}
+                        >
+                            Upload Proof of Income
+                        </FormButton>
+                        <Button
+                            variant="ghost"
+                            as="u"
+                            onClick={() => onNext()}
```
MatoPlus:
```suggestion
                            onClick={onNext}
```

src/components/registration-form/ParentEnrolledConfirmationPage.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { createClassRegistration } from "@utils/createClassRegistration";
+import { Student } from "@prisma/client";
+import { Loading } from "@components/Loading";
+
+type ParentEnrolledConfirmationPageProps = {
+    student: Student;
+    classId: number;
+};
+
+export const ParentEnrolledConfirmationPage: React.FC<ParentEnrolledConfirmationPageProps> =
```
MatoPlus, line 15: I'm guessing this page renders twice which is probably propagated somewhere up, causing the API call to be called twice. Usually, we want to do POST calls on a button press.
I can look into this later...

src/components/registration-form/DiscountPage.tsx
```diff
@@ -0,0 +1,74 @@
...
+                        your total.
+                    </Text>
+                    <Heading fontSize="22px" fontWeight="900">
+                        Has your income changed?
+                    </Heading>
+                    <Button
+                        variant="ghost"
+                        as="u"
+                        onClick={() => {
+                            setUpdateProofOfIncome(true);
```
MatoPlus: I'm not too sure what this is suppose to do... 
I think this should just redirect for document upload like in the submit proof of income page. After submitting the proof of income, we can mutate the me query and it will automatically update the the right page.

Edit: Oh or even better, just router.push("/myaccounts")

pages/parent/enrollment.tsx
```diff
@@ -77,6 +80,12 @@ export default function ParentEnrollClass({
         <MediaReleaseForm onNext={nextPage} />,
         <ParticipantWaiver onNext={nextPage} />,
         <TermsAndConditions onNext={nextPage} />,
+        <ProofOfIncomeWrapper
```
MatoPlus: Instead of passing a wrapper here, we can use a ternary on the high level To see if need the proof of income page, discount page, or nothing.

Kinda like what Brandon did here: https://github.com/uwblueprint/social-diversity-for-children/pull/107/files#diff-f65ce9fe6779f3e25edf7295e71921c959fa8fe6bb286360b0868c27c3a9658aR81-R109

MatoPlus (changes requested):
Looks like it's coming together! I tried to address your concerns. Two big things is to use the redirect query param on the document upload page properly and to conditionally set the page after terms and condition like how Brandon did in his PR.

I'll look at this PR more tomorrow, good work!

pages/parent/enrollment.tsx
```diff
@@ -23,10 +25,14 @@ type ParentEnrollClassProps = {
...
     session,
 }: ParentEnrollClassProps): JSX.Element {
-    const [pageNum, setPageNum] = useState<number>(0);
     const [selectedChild, setSelectedChild] = useState<number>(0);
     const { user, isLoading, error } = useUser(session.id as string);
     const router = useRouter();
+    const { classId, page } = router.query;
+    const numberClassId =
+        classId && classId.length > 0 ? Number(classId) : null;
+    const numberPage = page ? parseInt(String(page), 10) : null;
```
MatoPlus: nit, let's stay consistent with constructor vs `as`
```suggestion
        classId && classId.length > 0 ? classId as number : null;
    const numberPage = page ? parseInt(page as string, 10) : null;
```

pages/parent/enrollment.tsx
```diff
@@ -77,6 +83,20 @@ export default function ParentEnrollClass({
...
+                    classId={numberClassId}
+                    onNext={nextPage}
+                />
+            ) : user.parent.isLowIncome ? (
+                <DiscountPage onNext={nextPage} />
+            ) : (
+                <>{pageNum == 5 && nextPage()}</>
+            )}{" "}
+        </>,
     ];
```
MatoPlus: I was thinking we can simplify this a bit. That last page looks really sketchy.
```suggestion
    ];

    if (user.parent.isLowIncome) {
        pageElements.push(<DiscountPage onNext={nextPage} />);
    } else if (user.parent.proofOfIncomeLink === null) {
        pageElements.push(
            <ProofOfIncomePage
                pageNum={pageNum}
                classId={numberClassId}
                onNext={nextPage}
            />,
        );
    }
```

src/components/registration-form/DiscountPage.tsx
```diff
@@ -0,0 +1,67 @@
...
+                        as="u"
+                        onClick={() => {
+                            router.push("/myaccounts");
+                        }}
+                        alignSelf="flex-start"
+                        alignItems="flex-start"
+                        paddingLeft="0"
+                    >
+                        Update income
+                    </Button>
```
MatoPlus: Let's clean this button up a bit:
```suggestion
                    <Button
                        variant="link"
                        color="black"
                        fontWeight={400}
                        _hover={{ color: colourTheme.colors.Gray }}
                        onClick={() => {
                            router.push("/myaccounts");
                        }}
                        alignSelf="flex-start"
                        alignItems="flex-start"
                        paddingLeft="0"
                    >
                        <Text as="u">Update income</Text>
                    </Button>
```

src/components/registration-form/ProofOfIncomePage.tsx
```diff
@@ -0,0 +1,84 @@
...
+                        Upload Proof of Income
+                    </FormButton>
+                    <Button
+                        variant="ghost"
+                        as="u"
+                        onClick={onNext}
+                        borderRadius="6px"
+                    >
+                        Skip for Now
+                    </Button>
```
MatoPlus: Can you also help me out and clean up these skip buttons as well. I feel like this works a lot better:
```suggestion
                    <Button
                        variant="link"
                        color="black"
                        fontWeight={400}
                        _hover={{ color: colourTheme.colors.Gray }}
                        onClick={onNext}
                        borderRadius="6px"
                    >
                        <Text as="u">Skip for Now</Text>
                    </Button>
```

utils/createClassRegistration.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * createClassRegistration creates a registration for a class given a student
+ * @param student student's record to create
+ * @param classId id of class to create the record to
+ * @returns deleted records
+ */
+export async function createClassRegistration(
+    student: Student,
+    classId: number,
+) {
```
MatoPlus: Let's be explicit here with types
```suggestion
): Promise<any> {
```

src/components/registration-form/ParentEnrolledConfirmationPage.tsx
```diff
@@ -0,0 +1,70 @@
...
+    ({ student, classId }): JSX.Element => {
+        const data = createClassRegistration(student, classId);
+
+        if (!data) return <Loading></Loading>;
```
MatoPlus: I think I this should fix the double POST call issue. Always expect a page to render more than once, so we should always place safeguards.
```suggestion
        const [success, setSuccess] = useState(false);
        let data;

        if (!success) {
            data = createClassRegistration(student, classId).then((res) => res);
            setSuccess(true);
        }

        if (data === null) return <Loading />;
```

MatoPlus (changes requested):
I think just a couples more concerns and this is ready to merge. Everything should work after the suggestions below. The document page redirect to form might not work because of a small bug in the document upload page (a rebase on staging will fix this).

Can you help me edit `api/enroll/child` on line 129 and 140 to actually return a response?
```typescript
            // create parent registration record and return if it could not be created
            const newRegistration = await createParentRegistration(
                parentRegistrationInput,
            );
            if (!newRegistration) {
                ResponseUtil.returnBadRequest(
                    res,
                    `Registration could not be created`,
                );
                return;
            }

            ResponseUtil.returnOK(res, newRegistration);
            break;
```

This is optional for you but these pages should be translated, 
image
image
image

If you want to do this real quick, you would load the translations in the enrollment.tsx page under getServerProps function like:
```typescript
    return {
        props: {
            session,
            ...(await serverSideTranslations(context.locale, ["common", "form"])),
        },
    };
```

All the translations you need will be in "form.*" and "poi.*". Again, optional, could be cool to get use to. I can do this after this merges in otherwise!

MatoPlus:
@ZhangCreations Please upload a quick video if possible!

pages/parent/enrollment.tsx
ZhangCreations: Yeah, I mentioned that I thought it was sketchy in the PR notes. Thanks for this fix!

pages/parent/enrollment.tsx
ZhangCreations: Sounds good for the `numberPage`, however, for the `classId` it seems like it runs into a type error when converting the type since it thinks the parameter is of type `string | string[]`

MatoPlus: oh my bad! I guess we can do this for classId as well just to be consistent:
```typescript
    const numberClassId = classId ? parseInt(classId as string, 10) : null;
```

Sorry bout the nitpick. While we are at it, maybe we can do a `router.back()` if classId is null since it doesn't make sense to continue in that case.

MatoPlus (approved):
Ship it 🚢 

Thanks for much for getting this done man! I'll add the translations for the pages real quick after this gets in 👍

bekahma (approved):
nothing to add from design :3

PR #109, opened 2021-10-19
fix(seed): non-conflicting uid

MatoPlus:
### Notion ticket link

[Seeding for External MVP Fix](https://www.notion.so/uwblueprintexecs/Seeding-for-External-MVP-Fix-4959139a85884a9688320e504ca6d003)


MatoPlus:
Tested locally and on staging, the seeded ids were blocking the first X users from registering successfully. Merging quickly due to urgency.

PR #110, opened 2021-10-22
feat(s3-retrieve-file): add get s3 + cleanup

MatoPlus:
### Notion ticket link

Sample link workflow
[Rework file upload names + retrieve file](https://www.notion.so/uwblueprintexecs/Rework-file-upload-names-retrieve-file-442043c01a804a0ca0f6a60a913e8ae8)

https://user-images.githubusercontent.com/44826218/138379709-324c15ab-3890-40f0-9817-0e6b5894b890.mp4

- Use signed url to get items from bucket
- Add SWR hook for retrieve file
- Cleanup SWR hooks
- Add appropriate utils
- Reworked file upload paths to be `/<type>/<user-email>/filename.type` - eg: `criminal-check/rickson@uwbp.com/poi.pdf`

### Steps to test

1. Create a stub open file button
2. The right file should be downloaded, depending on whether or not you are a volunteer, parent

- Code style
- Any missing small oversights

### Checklist

-   [X] I have run the linter

### DIFF
--- a/pages/api/file/retrieve/index.ts
+++ b/pages/api/file/retrieve/index.ts
@@ -1,0 +1,81 @@
+import { getSignedUrlForRetrieve } from "@aws/s3";
+import { ResponseUtil } from "@utils/responseUtil";
+import { getSession } from "next-auth/client"; // Session handling
+import { NextApiRequest, NextApiResponse } from "next";
+import { roles } from ".prisma/client";
+import { getUserFromEmail } from "@database/user";
+
+export default async function handle(
+    req: NextApiRequest,
+    res: NextApiResponse,
+): Promise<void> {
+    const session = await getSession({ req });
+    // If there is no session
+    if (!session) {
+        return ResponseUtil.returnUnauthorized(res, "Unauthorized");
+    }
+
+    // TODO make this more robost/better
+    const accepted_type_paths = [
+        "criminal-check",
+        "income-proof",
+        "curriculum-plans",
+    ];
+
+    // Attempt to retrieve file requested in user namespace if possible
+    let { file, path } = req.query;
+
+    if (path && !accepted_type_paths.includes(path as string)) {
+        return ResponseUtil.returnNotFound(res, "Type not accepted");
+    }
+
+    switch (req.method) {
+        case "GET": {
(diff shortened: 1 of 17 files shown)


pages/api/file/retrieve/index.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    switch (req.method) {
+        case "GET": {
+            const user = await getUserFromEmail(session.user.email);
+
+            if (!user) {
+                return ResponseUtil.returnUnauthorized(res, "Unauthorized");
+            }
+            if (!path) {
```
ZhangCreations, line 39: Is there a specific reason we're defaulting to these values if the path is not specified

utils/fetcher.ts
```diff
@@ -3,6 +3,40 @@
...
+    let endpoint = `${url}`;
+
+    if (path && file) {
+        endpoint += `?path=${path}&file=${file}`;
+    } else if (path) {
+        endpoint += `?path=${path}`;
+    } else if (file) {
+        endpoint += `?file=${file}`;
+    }
+    console.log(endpoint);
```
ZhangCreations: remove?

ZhangCreations:
Took a quick look, will review more in depth at a later date. Thanks for bringing my attention to this code, it'll be super useful for my ticket.

pages/api/file/retrieve/index.ts
MatoPlus: Glad you brought this up, I feel like it's easier on the frontend, to do `/api/file/retrieve` rather than `/api/file/retrieve?path=criminal-check&file=somefile.pdf` right? I feel like this is a sensible default for the developers (at least for external), if the current is a volunteer, chances are, they will only retrieve their own criminal check rather than a non-existing income proof, vice versa for parent.

ie: look at how clean `useFileRetrieve` is 😄

utils/fetcher.ts
MatoPlus: yee

MatoPlus:
I'm going to merge this for now since we might need this our my accounts page. lmk if there are any concerns!

PR #116, opened 2021-10-26
fix(cleanup): hook and return bad request

MatoPlus:


MatoPlus:
Small cleanup merge for MVP

PR #118, opened 2021-10-27
draft - waitlist feature

aamyli:
### Notion ticket link

-

### Steps to test

1.

-

### Checklist

-   [ ] I have run the linter

### DIFF
--- a/src/components/ClassInfoCard.tsx
+++ b/src/components/ClassInfoCard.tsx
@@ -39,6 +39,9 @@ export const ClassInfoCard: React.FC<ClassInfoProps> = ({
     const { t } = useTranslation();
     const { me } = useMe();
 
+    const fullClass =
+        me && me.role === roles.PARENT && cardInfo.spaceAvailable < 1;
+
     return (
         <Grid
             templateColumns="repeat(4, 1fr)"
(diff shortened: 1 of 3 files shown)


src/components/ClassInfoCard.tsx
```diff
@@ -39,6 +39,9 @@ export const ClassInfoCard: React.FC<ClassInfoProps> = ({
     const { t } = useTranslation();
     const { me } = useMe();
 
+    const fullClass =
+        me && me.role === roles.PARENT && cardInfo.spaceAvailable < 1;
+
```
MatoPlus, line 44: hmm, although it's a bit ugly, I would think passing a new prop like isFull from class list -> class info card etc would be better then duplicating  this logic in your two `ClassInfoCard.tsx` and `ClassInfoModal.tsx`

MatoPlus (commented):
🔥 The WIP is looking good! If the waitlist creation is successful and the designers are happy with the UI, all that's left is the notify all on register logic, (I would think the hardest part would be the mailing). There will be other small things like deleting waitlist records after a parent successfully registers for a waitlisted class but honestly it can be another ticket if you wanna leave this and tackle other tickets.

PR #119, opened 2021-10-29
feat(admin): admin internal dashboard

MatoPlus:
### Notion ticket link

[Main Internal Dashboard Page](https://www.notion.so/uwblueprintexecs/Main-Internal-Dashboard-Page-1fd119199fbb4b9ebd5ba4e11a79825b)

https://user-images.githubusercontent.com/44826218/139561143-609cff07-7ef9-4a23-a58e-879d4ac2c88e.mp4

- Add page component under `/admin/index.tsx`
- This admin component uses new SWR queries to get stats and upcoming classes (along with any live classes)
- Uses existing getZoomLink SWR query to do join class
- The admin page will redirect to `/no-access` if user is not teacher or admin
- The stats route will only allow users who are teacher or admin to access

### Steps to test

1. Try to access `/admin` while logged in as parent - you should be redirected to a 404 page
2. Try to access `/admin` while logging teacher or admin - all stats should load correctly, upcoming and live classes are accurate, you can join the current live class

- Dev: Logic of get upcoming class backend is a bit janky especially for weekends, I think it's enough of an edge case but I'm open to ideas! 
- Design: Slight change to live class card - add volunteer count info (I feel like this would be important to admins)
- Design: I feel like the page is a bit squished, the dashboard header has not been added but when it does, we'll see how it goes 😅 . It will be taken care of in https://www.notion.so/uwblueprintexecs/Add-Dashboard-header-to-Admin-page-wrapper-211e9fa33291479dab24390078479e50 

### Checklist

-   [X] I have run the linter


MatoPlus:
@vedantdsn @bekahma lmk if you guys want me to create admin accounts for you guys to test this! It'll take a second!

gregmaxin:
> @vedantdsn @bekahma lmk if you guys want me to create admin accounts for you guys to test this! It'll take a second!

Can I get an admin or teacher account? Do I just change a boolean in the db? 

I also noticed that I get an error when logging in after clicking the link from the email (for parents/students) when I didn't before. I am not sure if related to the pr or not. 

I will review the code later

bekahma (approved):
the logic and everything looks great!! one small thing would be the margins between the metrics cards... if you can increase the spacing between the cards and decrease the margin between the # and the edge of the card 😄 def not high priority, just a small stylistic change for consistency

bekahma:
> @vedantdsn @bekahma lmk if you guys want me to create admin accounts for you guys to test this! It'll take a second!

yes that would be great :))

MatoPlus:
> > @vedantdsn @bekahma lmk if you guys want me to create admin accounts for you guys to test this! It'll take a second!
> 
> Can I get an admin or teacher account? Do I just change a boolean in the db?
> 
> I also noticed that I get an error when logging in after clicking the link from the email (for parents/students) when I didn't before. I am not sure if related to the pr or not.
> 
> I will review the code later

1. So how I did it was I just created a normal account and then go to the DB and assign the role column either the `PROGRAM_ADMIN` or the `TEACHER` role.
image. Feel free to create a PR to add yourself to the seed though! Hmu if you want some help.

2. Yea, it's as you say. The signup for these branch deployments are still not really working as it always redirect to the staging deployment for some reason. I thought I just needed to change the `NEXTAUTH_URL` env var but I might be missing something else...

MatoPlus:
> I also noticed that I get an error when logging in after clicking the link from the email (for parents/students) when I didn't before. I am not sure if related to the pr or not.

@GodGreg this has been fixed! It didn't redeploy after I changed the env var `NEXTAUT_URL=social-diversity-for-children-pr-119.up.railway.app`. It should work if you try again.

src/styles/colours.ts
```diff
@@ -5,12 +5,14 @@ const colourTheme = extendTheme({
         Blue: "#0C53A0",
         LightBlue: "#2C6AAD",
         Red: "#D8000C",
-        LightGrayBlue: "#E2E8F0",
+        CatskillWhite: "#E2E8F0",
+        LightGray: "#F8F8F8",
         Gray: "#737373",
+        PeriwinkleGray: "#DBE5F1",
+        Sliver: "#C5C5C5",
```
brandonwong368, line 12: typo!

utils/enum/weekday.ts
```diff
@@ -0,0 +1,83 @@
...
+        case weekday.WED:
+            return 3;
+        case weekday.THU:
+            return 4;
+        case weekday.FRI:
+            return 5;
+        case weekday.SAT:
+            return 6;
+        case weekday.SUN:
+            return 7;
```
gregmaxin, line 79: In `dateToWeekday` you have SUN -> SAT and in weekdayToDay you have MON -> SUN. If this works, great! Just wanted to catch any errors here if there is one

utils/hooks/useStats.ts
```diff
@@ -0,0 +1,28 @@
...
+    mutate: (data?: any, shouldRevalidate?: boolean) => Promise<any>;
+};
+/**
+ * Use hook for getting stats for admin dashboard
+ * @returns UseStatsResponse
+ */
+export default function useStats(): UseStatsResponse {
+    const { data, error, mutate } = useSWR("/api/admin/stats", fetcher);
+    return {
+        totalRegistrants: data && data.data ? data.data.totalRegistrants : null,
```
gregmaxin: Thoughts - no action required

I think you can just do `data?.data?.totalRegistrants` in ts instead of `data && data.data ? data.data.totalRegistrants : null`

Additionally, you could map the data to an array and then use the `...mappedData` notation before isLoading.

gregmaxin (approved):
Overall looks good to me. No major concerns in the code from me (other than what has already been commented). 

I have not done a thorough test of the UI since it seems someone else has already has.

utils/enum/weekday.ts
MatoPlus: yes, this was intentional. This new until is really for enum value comparisons, unlike the other converter which is for typing. Lmk if you know an easier way to do this.

utils/hooks/useStats.ts
MatoPlus: yes facts, that's a really good point. I'll make that change!

PR #124, opened 2021-11-03
feat(admin-view-class): view class page

MatoPlus:
### Notion ticket link

[View class participants](https://www.notion.so/uwblueprintexecs/View-class-participants-1e9bb234b1744ef8b14685986d6cd59d)

### Implementation description

https://user-images.githubusercontent.com/44826218/140614423-278868bf-dc43-4196-82c4-4652d0f2b65c.mp4

- Used breadcrumb for trail at the top
- Kinda stubby class card drop down option for now (no modal options, @brandonwong368 should have an implementation to integrate)
- Two tabs, student/volunteer
- Use new library, `react-table` for sorted table and global filter
- use new library `react-csv` for csv export
- private get class registrants for admins and teachers only (TODO)

### Steps to test

1. Test flow for class with volunteers
2. Test flow for class with students
3. Global filter filters any column
4. Export csv works and the output makes sense

### What should reviewers focus on?

- Dev: Code logic
- Dev: Any induced bugs?
- Design: Look and feel

### Checklist

-   [X] My PR name is descriptive and in imperative tense
-   [X] I have run the linter
-   [X] I have requested a review from the PL, and/or other devs who have background knowledge on this PR or who will be building on top of this PR

PR #126, opened 2021-11-05
Created Criminal Check Admin Page

ZhangCreations:
### Notion ticket link

[Criminal Record Check approval](https://www.notion.so/uwblueprintexecs/Criminal-Record-Check-approval-2fbfd762ac8e400c9ff1f4aa5ebdf28c)

- Implemented according to the design on Figma
- We don't have a date modified field so that line has been removed from the file
- Added null state for approval since default should be pending, not declined

### Steps to test

1. Go to the admin/criminal-check page and play around, most of the buttons do nothing except for the approval button on the file card, which should change the approval state on click.

- Check to see that the singular mutation works properly
- File downloads correctly

### Checklist

-   [x] I have run the linter

image


pages/admin/criminal-check.tsx
```diff
@@ -0,0 +1,177 @@
...
+                        Registrants
+                    </Text>
+                    <Button
+                        backgroundColor="white"
+                        borderColor="black"
+                        borderWidth="2px"
+                    >
+                        Add New Registrant
+                    </Button>
+                </HStack>
```
MatoPlus: This looks nice! We should make this a generic component. It should take in a title + an optional list of labels and hrefs as props. No action is needed for this PR if you don't want to. :P

pages/admin/criminal-check.tsx
```diff
@@ -0,0 +1,177 @@
...
+    session: Record<string, unknown>;
+};
+
+export default function CriminalCheck(props: AdminProps): JSX.Element {
+    const testData = {
+        volunteerName: "Test Name",
+        volunteerEmail: "kevinz999@gmail.com",
+        criminalRecordCheckLink: "blank.pdf",
+        criminalCheckApproved: false,
+    };
```
MatoPlus: nice. Ideally, we should replace this now with a useUser query in the future. I was thinking the criminal-check component would be a a user profile dynamic route such as `/admin/registrant/user/[id].tsx`. We would then use the route id to fetch info about the parent or volunteer.

pages/admin/criminal-check.tsx
```diff
@@ -0,0 +1,177 @@
...
+                            backgroundColor={colourTheme.colors.LightGray}
+                            display="flex"
+                            alignItems="center"
+                            justifyContent="center"
+                            textAlign="center"
+                            color={colourTheme.colors.Gray}
+                        >
+                            The participant has not uploaded a criminal record
+                            check at this time.
+                        </Box>
```
MatoPlus: We actually have a empty state component!
```suggestion
                        <EmptyState height="200px">
                            The participant has not uploaded a criminal record
                            check at this time.
                        </EmptyState>
```

To make it accept props, can you change the `/component/EmptyState.tsx` file to the following:
```typescript
import React from "react";
import { Box, Center, Flex, FlexProps } from "@chakra-ui/react";
import colourTheme from "@styles/colours";

type EmptyStateProps = FlexProps & {
    styleProps?: Record<string, unknown>;
};

export const EmptyState: React.FC<EmptyStateProps> = ({
    children,
    ...props
}): JSX.Element => {
    return (
        <Flex
            bg={colourTheme.colors.LightGray}
            w="100%"
            py={"40px"}
            px={"64px"}
            mt={5}
            display="flex"
            alignItems="center"
            justifyContent="center"
            textAlign="center"
            color={colourTheme.colors.Gray}
            {...props}
        >
            <Box>
                <Center whiteSpace="pre-line" textAlign="center">
                    {children}
                </Center>
            </Box>
        </Flex>
    );
};

```

pages/api/admin/update-criminal-check-status.ts
```diff
@@ -0,0 +1,49 @@
...
+export default async function handle(
+    req: NextApiRequest,
+    res: NextApiResponse,
+): Promise<void> {
+    const session = await getSession({ req });
+
+    // If there is no session or the user is not a parent, not authorized
+    if (
+        !session ||
+        ![roles.PROGRAM_ADMIN, roles.TEACHER].includes((session as any).role)
```
MatoPlus, line 21: I think it'd make more sense if we only allow program admins to approve criminal checks
```suggestion
        ![roles.PROGRAM_ADMIN].includes((session as any).role)
```

pages/api/admin/update-criminal-check-status.ts
```diff
@@ -0,0 +1,49 @@
...
+    ) {
+        return ResponseUtil.returnUnauthorized(res, "Unauthorized");
+    }
+
+    switch (req.method) {
+        case "POST": {
+            const input = req.body;
+            const volunteerEmail = input.email as string;
+            const criminalCheckApproval = input.approval as boolean | null;
```
MatoPlus: I'm not too sure what the intention is here?

prisma/schema.prisma
```diff
@@ -110,7 +110,7 @@ model Volunteer {
   dateOfBirth             DateTime       @map("date_of_birth") @db.Timestamptz(6)
   addressLine1            String?        @map("address_line1")
   criminalRecordCheckLink String?        @map("criminal_record_check_link")
-  criminalCheckApproved   Boolean?       @default(false) @map("criminal_check_approved")
+  criminalCheckApproved   Boolean?       @map("criminal_check_approved")
```
MatoPlus, line 114: good catch!

utils/updateFileApproval.ts
```diff
@@ -0,0 +1,36 @@
...
+        body: JSON.stringify(updateData),
+    };
+    console.log(request.body);
+
+    const response = await fetch(
+        `/api/admin/update-${filePath}-status`,
+        request,
+    );
+
+    mutate(`/api/admin/update-${filePath}-status`);
```
MatoPlus: here, the intention of the mutate is inform a SWR hook that it's data has been mutated and needs to be refetched.

In our case, /api/user/[id]` route...
Either way this will need to be tested later.

```suggestion
    mutate(`/api/user/${id}`);
```

utils/updateFileApproval.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * updateFileApproval updates a file's approval from an admin
+ * @param filePath
+ * @param email
+ * @param approval
+ * @returns deleted records
+ */
+export async function updateFileApproval(
+    filePath: string,
+    email: string,
```
MatoPlus: I think it'd be easier for us to use ids instead since we are playing to use dynamic routes for these registrant pages via their user ids.
```suggestion
    id: number, 
```

we'll need to change the logic below accordingly.

src/components/fileDownloadCard.tsx
```diff
@@ -0,0 +1,129 @@
...
+    docApproved: boolean | null;
+    participantEmail: string;
+};
+
+const FileDownloadCard: React.FC<FileDownloadCardProps> = ({
+    filePath,
+    docLink,
+    docName,
+    docApproved,
+    participantEmail,
```
MatoPlus: idem, we want user id instead of email here due to the dynamic route.

src/components/fileDownloadCard.tsx
```diff
@@ -0,0 +1,129 @@
...
+                    as={MdDescription}
+                    color={colourTheme.colors.Blue}
+                    w={110}
+                    h={110}
+                    marginLeft="40px"
+                    marginTop="40px"
+                />
+                <VStack alignItems="flex-start" w="300px" h="full">
+                    <Text fontSize="18px" marginTop="75px" as="u">
+                        <Link href={docLink} color={colourTheme.colors.Blue}>
```
MatoPlus: I think we need something like an isExternal prop for this to open in another tab?
```suggestion
                        <Link href={docLink} color={colourTheme.colors.Blue} isExternal>
```

MatoPlus (changes requested):
Looks like the frontend logic is coming together, great work!

For the criminal check page, I imagine this will eventually become a part of `/admin/registrant/user/[id].tsx` Can you make the page such that it accepts an id for a volunteer and properly fetch their files?

I think I didn't integrate admin access into the `/api/file/retrieve`  handler yet. Can you add a bypass for admins? Otherwise, the useFileRetrieve hook you're using would not work internally since it currently assumes that you are fetching your own files (which is not the case for admins). This will require some backend api tweaking!

I think this will have quite some overlap with this ticket: https://www.notion.so/uwblueprintexecs/View-edit-registrant-ee9869a2bb8049c8b4b33046e96d1823 so it would be awesome if you can cover that ticket for the sprint as well! This is definitely an important page that I think you can finish 💪

bekahma (commented):
couple of comments from design:
1. i know this is not mocked up in the figma, but is it possible to create a default state for the badge? something like `new`? this will be before admin takes it into `pending`.
2. i assume the date submitted field will be added later bc we're waiting on external correct?

everything looks great otherwise!

MatoPlus:
> couple of comments from design:
> 
> 1. i know this is not mocked up in the figma, but is it possible to create a default state for the badge? something like `new`? this will be before admin takes it into `pending`.
> 2. i assume the date submitted field will be added later bc we're waiting on external correct?
> 
> everything looks great otherwise!

@bekahma I think this is a bit of extra work atm since it requries us to change the schema. I think currently, since we are only ever showing 1 criminal check, three states should be fine! Lmk you thoughts. I can definitely make a ticket for it for our backlog.

2. The submitted date will be added in the soon. We already have a ticket in progress. I was thinking it might be a cool ticket for @ZhangCreations to tackle! Here's the ticket: https://www.notion.so/uwblueprintexecs/Add-Submit-Date-Columns-to-bgc-poi-for-myaccounts-b652d1a6dad14b70bf99432ddc152cf9 lmk if you wanna sync on this.

pages/api/admin/update-criminal-check-status.ts
MatoPlus: @ZhangCreations Is this functioning as a bitwise operator? Was there a problem with just doing `const criminalCheckApproval = input.approval as boolean`

pages/api/admin/update-criminal-check-status.ts
```diff
@@ -0,0 +1,48 @@
...
+        ![roles.PROGRAM_ADMIN, roles.TEACHER].includes((session as any).role)
+    ) {
+        return ResponseUtil.returnUnauthorized(res, "Unauthorized");
+    }
+
+    switch (req.method) {
+        case "POST": {
+            const input = req.body;
+            const volunteerId = input.id as number;
+            const criminalCheckApproval = input.approval as boolean | null;
```
MatoPlus: @ZhangCreations Is this functioning as a bitwise operator? Was there a problem with just doing
```suggestion
            const criminalCheckApproval = input.approval as boolean;
```

src/components/EmptyState.tsx
```diff
@@ -1,28 +1,36 @@
...
 }): JSX.Element => {
     return (
-        <Box
+        <Flex
             bg={colourTheme.colors.LightGray}
             w="100%"
             py={"40px"}
             px={"64px"}
             mt={5}
+            display="flex"
```
MatoPlus: Sorry mb, I think this line is redundant
```suggestion
```

utils/updateFileApproval.ts
```diff
@@ -0,0 +1,35 @@
...
+    };
+
+    const request = {
+        method: "POST",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify(updateData),
+    };
+
+    const response = await fetch(
+        `/api/admin/update-${filePath}-status`,
```
MatoPlus, line 27: @ZhangCreations  hmm this is an interesting way of doing things. Do you think it'd make sense to add one more level of abstraction here and make a updateCriminalCheckApproval, updateIncomeProofApproval, etc... which uses this util? That way you don't have to keep a random FILE_PATH const in the main page.  Keeping stuff contained 👍 

So, instead of using updateFileApproval in your page, you'll just use updateCriminalCheckApproval instead, which calls this util indirectly.

MatoPlus (changes requested):
Starting to look good!

Also tip, for the breadcrumb at the top:
image

please use the chakra ui breadcrumb component to save yourself some work!
https://chakra-ui.com/docs/navigation/breadcrumb

For example: https://github.com/uwblueprint/social-diversity-for-children/pull/124/files#diff-1b78b08df8e3f1925f16f68fb82f03d6bd29ff555290d2b16f15531532e14d3fR68-R88

utils/updateFileApproval.ts
ZhangCreations: I think I personally like the modularity of this component more, it makes less files and maintaining the FILE_PATH const in the top level isn't too much additional data. Up to you however, let me know if I should change this.

pages/api/admin/update-criminal-check-status.ts
MatoPlus: oh nvm, It's `const criminalCheckApproval = input.approval as (boolean | null);` right? I read it as `  const criminalCheckApproval = (input.approval as boolean) | null;` for some reason lol. Looks good to me!

utils/updateFileApproval.ts
MatoPlus: I like this util as well, it's clean. It's just that stating the FILE_PATH through multiple pages could still be redundant and I'd rather call `updateCriminalCheckApproval(id, true)` vs something like `updateFileApproval("criminal-check", id, true)` everytime. I was thinking we can make another function that we can export in this file. Something like


```typescript
export async function updateCriminalCheckApproval(
    id: number,
    approval: boolean | null,
): Promise<Response> {
    return updateFileApproval("criminal-check", id, approval)
}
```

it's not much additional code and it goes a long way 😄 ,

ZhangCreations: If I make a function `updateCriminalCheckApproval` for the fileDownload card it makes it so that the fileDownloadCard component is no longer reusable for the proof of income page.

MatoPlus: That's true. My judgement is that we probably don't want to be passing slightly arbitrary strings around. Maybe we can even create an enum of some kind then I'd be happy with passing the enum around instead. Maybe for now, do you think It'd make sense to specify the type of file path as `"criminal-check" | "income-proof"`?

bekahma:
oh i forgot about this until now - did we account for the empty state in this PR?

MatoPlus:
> oh i forgot about this until now - did we account for the empty state in this PR?

Yup! It's in the code, not in the screenshot sadly :P

pages/admin/registrant/user/[id].tsx
```diff
@@ -0,0 +1,181 @@
...
+                        >
+                            Registrants
+                        </Text>
+                        <Button
+                            backgroundColor="white"
+                            borderColor="black"
+                            borderWidth="2px"
+                        >
+                            Add New Registrant
+                        </Button>
```
Comment (MatoPlus) [inline pages/admin/registrant/user/[id].tsx]: Design got updated, this is no longer needed
```suggestion
```

pages/admin/registrant/user/[id].tsx
```diff
@@ -0,0 +1,181 @@
+import React, { useState, useEffect } from "react";
```
Comment (MatoPlus) [inline pages/admin/registrant/user/[id].tsx]: looks like you're separating the profile pages by role. Do you mind changing the name of this page to `/admin/registrant/volunteer/[id].tsx`?

Ideally this page would support both volunteer and parent users but this is totally fine.

pages/admin/registrant/user/[id].tsx
```diff
@@ -0,0 +1,181 @@
...
+} from "@chakra-ui/react";
+import { ChevronRightIcon } from "@chakra-ui/icons";
+
+import colourTheme from "@styles/colours";
+import { MdPerson, MdDescription } from "react-icons/md";
+import FileDownloadCard from "@components/fileDownloadCard";
+import { Volunteer } from ".prisma/client";
+import EmptyState from "@components/EmptyState";
+
+const FILE_PATH = FileType.CRIMINAL_CHECK;
```
Comment (MatoPlus) [inline pages/admin/registrant/user/[id].tsx]: We don't need this now right?

```suggestion
```

pages/admin/registrant/user/[id].tsx
```diff
@@ -0,0 +1,181 @@
...
+                                color={colourTheme.colors.Blue}
+                                width="full"
+                            >
+                                <Icon as={MdDescription} w={8} h={8} />
+                                <Link>Criminal Record Check</Link>
+                            </HStack>
+                        </VStack>
+                        {volunteerData.criminalRecordCheckLink !== null ? (
+                            <FileDownloadCard
+                                filePath={FILE_PATH}
```
Comment (MatoPlus) [inline pages/admin/registrant/user/[id].tsx]: instead of using const, we can just use enum directly
```suggestion
                                filePath={FileType.CRIMINAL_CHECK}
```

pages/admin/registrant/user/[id].tsx
```diff
@@ -0,0 +1,181 @@
...
+    const {
+        user,
+        isLoading: userIsLoading,
+        error: userError,
+    } = useUser(userId as string);
+
+    if (userIsLoading) {
+        return <Loading />;
+    }
+    if (user) {
```
Comment (MatoPlus) [inline pages/admin/registrant/user/[id].tsx]: I feel like we don't need a if statement here since the loading case is already handled above.
```suggestion
```

pages/admin/registrant/user/[id].tsx
```diff
@@ -0,0 +1,181 @@
...
+                                The participant has not uploaded a criminal
+                                record check at this time.
+                            </EmptyState>
+                        )}
+                    </HStack>
+                </VStack>
+            </Wrapper>
+        );
+    }
+    return <Loading />;
```
Comment (MatoPlus) [inline pages/admin/registrant/user/[id].tsx]: idem
```suggestion
```

pages/admin/registrant/user/[id].tsx
```diff
@@ -0,0 +1,181 @@
...
+ * getServerSideProps gets the session before this page is rendered
+ */
+export const getServerSideProps: GetServerSideProps = async (context) => {
+    // obtain the next auth session
+    const session = await getSession(context);
+
+    return {
+        props: {
+            session,
+            ...(await serverSideTranslations(context.locale, ["content"])),
```
Comment (MatoPlus) [inline pages/admin/registrant/user/[id].tsx]: no translations for internal needed
```suggestion
```

pages/admin/registrant/user/[id].tsx
```diff
@@ -0,0 +1,181 @@
...
+    return <Loading />;
+}
+
+/**
+ * getServerSideProps gets the session before this page is rendered
+ */
+export const getServerSideProps: GetServerSideProps = async (context) => {
+    // obtain the next auth session
+    const session = await getSession(context);
+
```
Comment (MatoPlus) [inline pages/admin/registrant/user/[id].tsx:191]: kick user if they are not admin

```suggestion
    if (!session) {
        return {
            redirect: {
                destination: "/login",
                permanent: false,
            },
        };
    } else if (!isAdmin(session)) {
        return {
            redirect: {
                destination: "/no-access",
                permanent: false,
            },
        };
    }
```

import isAdmin via `import { isAdmin } from "@utils/session/authorization";`

pages/api/admin/update-criminal-check-status.ts
```diff
@@ -0,0 +1,48 @@
...
+    req: NextApiRequest,
+    res: NextApiResponse,
+): Promise<void> {
+    const session = await getSession({ req });
+
+    // If there is no session or the user is not a parent, not authorized
+    if (
+        !session ||
+        ![roles.PROGRAM_ADMIN, roles.TEACHER].includes((session as any).role)
+    ) {
```
MatoPlus: Let's simplify this:
```suggestion
    if (!session || !isAdmin(session)) {
```

MatoPlus (changes requested):
Just a couple more items!

MatoPlus:
I DMed about the bad api/file/retrieve behavior. Depending on Kevin to get it fixed 🙏

pages/admin/registrant/user/[id].tsx
```diff
@@ -47,122 +44,137 @@ export default function CriminalCheck(props: AdminProps): JSX.Element {
 
     if (userIsLoading) {
         return <Loading />;
```
Comment (MatoPlus) [inline pages/admin/registrant/user/[id].tsx]: please use the AdminLoading component 🙏. look up on it's current usage. It's basically the loading component wrapped in the navbar

pages/admin/registrant/user/[id].tsx
```diff
@@ -47,122 +44,137 @@ export default function CriminalCheck(props: AdminProps): JSX.Element {
...
+    } else if (userError) {
+        router.push("/404").then(() => window.scrollTo(0, 0));
```
Comment (MatoPlus) [inline pages/admin/registrant/user/[id].tsx]: let's return the `AdminError` component instead. Look up usage for reference. You'll need to pass in some cause like "cannot fetch user" or something. Test this by visiting a page that does not exists like: "/admin/user/99999`

utils/fetcher.ts
```diff
@@ -38,23 +38,20 @@ export function fetcherWithId(
...
-    } else if (file) {
-        endpoint += `?file=${file}`;
-    }
-
-    return fetcher(endpoint);
+    let endpoint = `${url}?`;
+    if (path) endpoint += `path=${path}&`;
+    if (file) endpoint += `file=${file}&`;
+    if (email) endpoint += `email=${encodeURIComponent(email)}&`;
+    console.log(encodeURIComponent(email));
```
MatoPlus: not needed?

MatoPlus (changes requested):
I'll take a closer look tonight, but please fix all the github action warnings you can!

pages/admin/registrant/user/[id].tsx
```diff
@@ -0,0 +1,211 @@
...
+        return <AdminError cause="registrants could not be loaded" />;
+    }
+
+    const userRole = user.role as string;
+    let volunteerData, parentData;
+    if (userRole == "VOLUNTEER") {
+        volunteerData = user?.volunteer as Volunteer;
+    } else if (userRole == "PARENT") {
+        parentData = user?.parent as Parent;
+    }
```
Comment (MatoPlus) [inline pages/admin/registrant/user/[id].tsx:59]: I like this if statement you got here : )
I think adding a else clause to return an `<AdminError cause="registrant not found" />;` or something along that line would be more robust, in case the id belongs to an internal user.

pages/admin/registrant/user/[id].tsx
```diff
@@ -0,0 +1,211 @@
...
+
+    if (userIsLoading) {
+        return <AdminLoading />;
+    } else if (userError) {
+        return <AdminError cause="registrants could not be loaded" />;
+    }
+
+    const userRole = user.role as string;
+    let volunteerData, parentData;
+    if (userRole == "VOLUNTEER") {
```
Comment (MatoPlus) [inline pages/admin/registrant/user/[id].tsx]: I think instead of string comparisons, we can actually compare enums.

```suggestion
    if (userRole == roles.VOLUNTEER) {
```
which you can do after roles from `@prisma/client `.

pages/admin/registrant/user/[id].tsx
```diff
@@ -0,0 +1,211 @@
...
+        volunteerData = user?.volunteer as Volunteer;
+    } else if (userRole == "PARENT") {
```
Comment (MatoPlus) [inline pages/admin/registrant/user/[id].tsx]: idem

pages/admin/registrant/user/[id].tsx
```diff
@@ -0,0 +1,211 @@
...
+                            <Icon as={MdDescription} w={8} h={8} />
+                            <Link>
+                                {" "}
+                                {userRole === "VOLUNTEER"
+                                    ? "Criminal Record Check"
+                                    : "Proof of Income"}
+                            </Link>
+                        </HStack>
+                    </VStack>
+                    {userRole === "VOLUNTEER" ? (
```
Comment (MatoPlus) [inline pages/admin/registrant/user/[id].tsx]: idem, use enum

pages/admin/registrant/user/[id].tsx
```diff
@@ -0,0 +1,211 @@
...
+                        </HStack>
+                        <HStack
+                            fontWeight={700}
+                            color={colourTheme.colors.Blue}
+                            width="full"
+                        >
+                            <Icon as={MdDescription} w={8} h={8} />
+                            <Link>
+                                {" "}
+                                {userRole === "VOLUNTEER"
```
Comment (MatoPlus) [inline pages/admin/registrant/user/[id].tsx]: idem, use enum

pages/admin/registrant/user/[id].tsx
```diff
@@ -0,0 +1,211 @@
...
+                        <BreadcrumbLink href="#">
+                            Browse Registrants
+                        </BreadcrumbLink>
+                    </BreadcrumbItem>
+                    <BreadcrumbItem>
+                        <BreadcrumbLink href="#">{userName}</BreadcrumbLink>
+                    </BreadcrumbItem>
+                    <BreadcrumbItem isCurrentPage>
+                        <BreadcrumbLink href="#">
+                            {userRole === "VOLUNTEER"
```
Comment (MatoPlus) [inline pages/admin/registrant/user/[id].tsx]: idem, use enum

prisma/schema.sql
```diff
@@ -92,6 +92,7 @@ CREATE TABLE parents (
   is_low_income BOOLEAN DEFAULT false,
   preferred_language locales NOT NULL,
   proof_of_income_link TEXT,
+  proof_of_income_approval BOOLEAN DEFAULT null,
```
MatoPlus: no action needed but do we need a default null? is it not implicit?

pages/api/file/retrieve/index.ts
```diff
@@ -59,7 +61,15 @@ export default async function handle(
                 }
             }
 
-            const uploadFilePath = `${path}/${user.email}/${file}`;
+            if (email && email !== user.email) {
+                if (!isAdmin(user)) {
+                    return ResponseUtil.returnUnauthorized(res, "Unauthorized");
+                }
+            }
```
MatoPlus, line 68: nice!

utils/fetcher.ts
```diff
@@ -38,23 +38,19 @@ export function fetcherWithId(
...
-    } else if (file) {
-        endpoint += `?file=${file}`;
-    }
-
-    return fetcher(endpoint);
+    let endpoint = `${url}?`;
+    if (path) endpoint += `path=${path}&`;
+    if (file) endpoint += `file=${file}&`;
+    if (email) endpoint += `email=${encodeURIComponent(email)}&`;
+    return fetcher(endpoint.slice(0, -1));
```
MatoPlus, line 55: thanks for smoothing this out!

MatoPlus (approved):
Lookin good @ZhangCreations! Appreciate the little cleanups as well : )

Just a couple of nits and you're free to merge this in!

PR #69, opened 2021-08-18
Scheduled DB Query for Email Cron Job

victoranyun:
### Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/Cron-job-to-send-reminder-emails-4f2a2a05adba489bbf75d66d7c69a362)

### Implementation description

- Set up AWS Lambda to ping backend every hour @ /api/mail
- Endpoint checks DB for any classes that are starting in 2-4 hours from the current time and 47-49 hours from the current time
- Sends email to all volunteers/parents registered for these classes

Screen Shot 2021-08-19 at 2 36 00 AM
Screen Shot 2021-08-19 at 2 36 46 AM
Screen Shot 2021-08-19 at 2 35 38 AM

Screen Shot 2021-08-29 at 6 31 56 PM

### Steps to test

1. Setup records in local DB
2. /api/mail

### What should reviewers focus on?

- Email template and what contents and how to abstract

### Checklist

-   [x] My PR name is descriptive and in imperative tense
-   [x] I have run the linter
-   [x] I have requested a review from the PL, and/or other devs who have background knowledge on this PR or who will be building on top of this PR

PR #130, opened 2021-11-08
send waitlist email flow

aamyli:
### Notion ticket link

-

### Steps to test

1.

-

### Checklist

-   [ ] I have run the linter


pages/api/waitlist/mail.ts
```diff
@@ -0,0 +1,3 @@
+// mail here ??
+// write something similar to pages/api/mail/index.ts?
+// and call when someone unregisters for a class?
```
aamyli, line 3: @MatoPlus was wondering if you think this is the correct directory to write the POST(?) request? then the endpoint would be api/waitlist/mail 

i see the "Class in X hours" email in pages/api/mail/index.ts but also that doesn't seem to be called anywhere (is it run automatically?) ; i was thinking of writing the request here but the main difference is that it'd have to be called (after the deleteClassRegistration)

utils/mail/templateWaitlist.ts
```diff
@@ -0,0 +1 @@
+// email ?
```
aamyli, line 1: + template for email would be here;

i think i'd pass in params for the class + its related data to be displayed in the email? i'm thinking the email can say something like

"X class currently has spots available" + provide some info (date, time of class etc)
-> from there, the email can have a button that links directly to the browse classes page? and button can say "Register Now" or something
-> we could also add another button that says "Remove from Waitlist" and link it to the MyClasses page?

@bekahma @vedantdsn @MatoPlus what do you think ^^ ; if you have any ideas for how you want the email to look like let me know! otherwise, i don't mind coding up the draft and we can make alterations after

pages/api/waitlist/mail.ts
MatoPlus: `/api/mail/index` is a hook that is called via a hourly cron job via our AWS lambda (serverless) function. This lambda function is actually in the production account if you want to take a look. The PR was here https://github.com/uwblueprint/social-diversity-for-children/pull/69

I was thinking after a deleteRegistration api request is called, we just directly call a mail promise push in the api logic to send mails in the backend.

Similarly to what is in the `/api/mail/index` page,  we should add something similar to the following lines:
```typescript
        (parentRegs as any[]).forEach((reg) => {
            mailerPromises.push(
                send(
                    process.env.EMAIL_FROM,
                    reg.parent.user.email,
                    getEmailSubject(reg.parent.preferredLanguage),
                    classStartingSoonTemplate(
                        intervalHours,
                        imageLink,
                        name,
                        weekday,
                        startDate,
                        endDate,
                        startTimeMinutes,
                        durationMinutes,
                        reg.parent.preferredLanguage,
                    ),
                ),
            );
        });
```
^^ /api/mail line 90

MatoPlus: @aamyli  Let me know if you want to sync up a bit about how notification mailing should work.

MatoPlus: Basically, we don't even want to create another post method. It's going to be a util method at most to be called in the deleteRegistration route.

utils/mail/templateWaitlist.ts
MatoPlus: That sounds good to me! Go for it 🔥

PR #131, opened 2021-11-08
Implement admin browse programs page

brandonwong368:
### Notion ticket link

[Admin Browse Programs Page](https://www.notion.so/uwblueprintexecs/6dd603d3687743ce8d09d6298dc99748?v=7332b6570bcc4b48b5cd73fa2719bec8&p=f8bfd5cc486a46999d9676d731b5156f)

- ticket is split into browse programs/browse classes. This PR is for browse programs page
- created new page /admin/programs to view all available programs
- displays all the programs through a card in a 3 column grid
- implemented a filter feature using the search bar to only display the program card if the name/description matches the filter query
- if no programs match the filter, an empty state is returned 

https://user-images.githubusercontent.com/71860803/141852436-6997d56e-104b-48b7-9bc3-3314e227b3d2.mp4

### Steps to test

1. Test that all the programs appear
2. Type into the search bar to filter through program names and their descriptions

### Checklist

-   [x] I have run the linter


pages/admin/programs/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+import { ReactNode } from "react";
+import colourTheme from "@styles/colours";
+import { SmallAddIcon } from "@chakra-ui/icons";
+import { SearchIcon } from "@chakra-ui/icons";
+import { BrowseProgramCard } from "@components/BrowseProgramCard";
+import { locale } from "@prisma/client";
+import usePrograms from "@utils/hooks/usePrograms";
+import { useRouter } from "next/router";
+import { Loading } from "@components/Loading";
+import { GetServerSideProps } from "next"; // Get server side props
```
MatoPlus:
```suggestion
import { GetServerSideProps } from "next";
```

pages/admin/programs/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+            <Button
+                fontSize="16px"
+                backgroundColor="white"
+                fontWeight="400"
+                px={"16px"}
+                py={"8px"}
+                border-radius="6px"
+                border="1px solid #0C53A0"
+            >
+                <SmallAddIcon mr="11px" />
```
MatoPlus: I think you can do something like using leftIcon prop for buttons.

pages/admin/programs/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+export default BrowsePrograms;
+
+/**
+ * getServerSideProps gets the session before this page is rendered
+ */
+export const getServerSideProps: GetServerSideProps = async (context) => {
+    // obtain the next auth session
+    const session = await getSession(context);
+    // TODO : check session for admin/volunteer , if not redirect to no access
```
MatoPlus: something like this works for now. I'll clean this up later with a helper method.

```suggestion
    if (!session) {
        return {
            redirect: {
                destination: "/login",
                permanent: false,
            },
        };
    } else if (
        // eslint-disable-next-line @typescript-eslint/no-explicit-any
        ![roles.PROGRAM_ADMIN, roles.TEACHER].includes((session as any).role)
    ) {
        return {
            redirect: {
                destination: "/no-access",
                permanent: false,
            },
        };
    }
```

pages/admin/programs/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+    const filteredCards = programCardInfos.filter((val) => {
+        if (searchTerm == "") {
+            return val;
+        } else if (
+            val.name.toLowerCase().includes(searchTerm.toLowerCase()) ||
+            val.description.toLowerCase().includes(searchTerm.toLowerCase())
+        ) {
+            return val;
+        }
+    });
```
MatoPlus: maybe make iterated name more clear + search tags
```suggestion
    const filteredCards = programCardInfos.filter((prog) => {
        const term = searchTerm.toLowerCase();
        if (term == "") {
            return prog;
        } else if (
            prog.name.toLowerCase().includes(term) ||
            prog.description.toLowerCase().includes(term) ||
            prog.onlineFormat.toLowerCase().includes(term) ||
            prog.tag.toLowerCase().includes(term)
        ) {
            return prog;
        }
    });
```

pages/admin/programs/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+                            ))}
+                        </HStack>
+                    </Flex>
+                </Box>
+                <Divider
+                    orientation="horizontal"
+                    marginTop="25px"
+                    marginBottom="30px"
+                    border="2px"
+                />
```
MatoPlus: Do you think we can make this a component? Since we'll be using this in almost every page. Also please fix this up a bit to make it look more like the header in our design (in terms of padding)!

pages/admin/programs/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+            val.description.toLowerCase().includes(searchTerm.toLowerCase())
+        ) {
+            return val;
+        }
+    });
+    return (
+        <Wrapper session={props.session}>
+            <Box>
+                <Box bg={"transparent"} px={"50px"} pt={"20px"} mx={"auto"}>
+                    <Flex h={"94px"}>
```
MatoPlus:
```suggestion
                    <Flex>
```
don't think height is needed

pages/admin/programs/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+                                setSearchTerm(e.target.value);
+                            }}
+                        />
+                    </InputGroup>
+                </Box>
+
+                <Box ml="50px" mt="25px">
+                    {filteredCards && filteredCards.length > 0 ? (
+                        <Grid templateColumns="repeat(3, 1fr)" gap={6}>
+                            {" "}
```
MatoPlus:
```suggestion
```

pages/admin/programs/index.tsx
MatoPlus: we can reduce the margin bit to 24px
```suggestion
                <Divider orientation="horizontal" my={6} border="2px" />
```

src/components/BrowseProgramCard.tsx
```diff
@@ -0,0 +1,104 @@
...
+                                children={cardInfo.tag}
+                            />
+                            <SDCBadge
+                                ml="2"
+                                backgroundColor={
+                                    colourTheme.colors.CatskillWhite
+                                }
+                                color={colourTheme.colors.Blue}
+                                children={cardInfo.onlineFormat}
+                            />
```
MatoPlus: use existing component:
```suggestion
                            <AdminBadge>{cardInfo.tag}</AdminBadge>
                            <AdminBadge ml={2}>
                                {cardInfo.onlineFormat}
                            </AdminBadge>
```

pages/admin/programs/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+export default BrowsePrograms;
+
+/**
+ * getServerSideProps gets the session before this page is rendered
+ */
+export const getServerSideProps: GetServerSideProps = async (context) => {
+    // obtain the next auth session
+    const session = await getSession(context);
+    // TODO : check session for admin/volunteer , if not redirect to no access
+    // refer to github
```
MatoPlus:
```suggestion
```

pages/admin/programs/[pid].tsx
```diff
@@ -0,0 +1,155 @@
...
+                border-radius="6px"
+                border="1px solid #0C53A0"
+            >
+                <SmallAddIcon mr="11px" />
+                {children}
+            </Button>
+        </HStack>
+    </Link>
+);
+
```
Comment (MatoPlus) [inline pages/admin/programs/[pid].tsx]: Add a TODO here

```suggestion
// TODO: Complete page
```

src/components/ProgramDescriptionCard.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { useRouter } from "next/router";
+import { useTranslation } from "react-i18next";
+import useMe from "@utils/hooks/useMe";
+
+type ProgramDescriptionCardProps = {
+    cardInfo: ClassCardInfo;
+};
+
+/**
+ *
```
MatoPlus: Add a TODO here

```suggestion
 * TODO: Complete page
```

MatoPlus (changes requested):
Looks good overall, just a couple of suggestions.

Please rebase so we can get the nice github action checks matt added.

bekahma (approved):
👍  from design

pages/admin/programs/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+                                    <GridItem key={idx}>
+                                        <BrowseProgramCard cardInfo={item} />
+                                    </GridItem>
+                                );
+                            })}
+                        </Grid>
+                    ) : (
+                        <Box pr="50px">
+                            <AdminEmptyState
+                                w="100"
```
MatoPlus: is this suppose to be 100%?
```suggestion
                                w="100%"
```

pages/admin/programs/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+                                h="100%"
+                                isLoading={isLoading}
+                            >
+                                No filtered classes available
```
MatoPlus: I think we also need to consider the case where we do not have any filtered classes, as in no programs. I think the following label might make more sense. lmk if you have any thoughts.

cc: @bekahma @bekahma 

```suggestion
                                No programs available
```

brandonwong368: @bekahma yup!! There wasn't an empty state design so I kind of improvised it

brandonwong368: It looks like this rn 
image

pages/admin/programs/index.tsx
```diff
@@ -0,0 +1,146 @@
+import {
```
MatoPlus: Can we make this the `program/index.tsx` page instead? That's what the navbar assumes and usually, we use singulars instead of plural for page routes because it doesn't give any semantic meaning. Ie: program -> all program, program/[id] -> singular program.

pages/admin/programs/index.tsx
MatoPlus: hmm, maybe we can make the height a bit bigger?

brandonwong368: approx how big?

MatoPlus: 2-2.5x current height maybe? eyeball it for look and feel. unless @uwblueprint/sdc-design has any opinions.

MatoPlus (approved):
LGTM

PR #137, opened 2021-11-15
feature - send waitlist emails when parent unregisters from class

aamyli:
### Notion ticket link

[Notion Ticket](https://www.notion.so/uwblueprintexecs/Send-email-after-class-registration-1e0c469cf3af466d812b9a84c2e161b3)
[Loom Demo Video](https://www.loom.com/share/b52c42d3accc483e91feaf01a55888f3)

Screen Shot 2021-11-14 at 7 29 16 PM

- in the DELETE request of `api/enroll/child` for a particular class, we:
1. get all waitlist records with same class ID
2. if there are records, we get the class info & for each class, we push an email to `mailerPromises`
3. then, send all waitlist emails 

### Steps to test

Test cases/things to consider:
- check if email will be sent to all (2+) waitlisted records for a specific class
-- I tested with 2 emails for one class waitlist, both emails received the waitlist email on unenrollment

- currently, a parent IS able to both be enrolled in a class & on the waitlist for the class (ie they may be waiting for another child)
- if multiple unenrollments occur within a time range, parents on waitlist would get multiple of the same email -> not sure if we want to do anything about this?

- email template, wording, etc.

### Checklist

-   [x] I have run the linter

### DIFF
--- a/pages/api/enroll/child.ts
+++ b/pages/api/enroll/child.ts
@@ -166,6 +166,23 @@ export default async function handle(
                 return;
             }
 
+            // check if the class is full, prior to deletion
+            let notifyWaitlist = true;
+            const waitlistClass = await getClass(
+                parentRegistrationInput.classId,
+            );
+            if (!waitlistClass) {
+                ResponseUtil.returnBadRequest(
+                    res,
+                    `Class information could not be found`,
+                );
+                return;
+            }
+            // if class wasn't originally full, do not notify waitlist
+            if (waitlistClass._count.parentRegs < waitlistClass.spaceTotal) {
+                notifyWaitlist = false;
+            }
+
             const deletedRegistration = await deleteParentRegistration(
                 parentRegistrationInput,
             );
@@ -182,23 +199,11 @@ export default async function handle(
                 parentRegistrationInput.classId,
             );
             // if there are, get class information & send email to all on waitlist
-            if (waitlistRecords.length > 0) {
-                const waitlistClass = await getClass(
-                    parentRegistrationInput.classId,
-                );
-                if (!waitlistClass) {
-                    ResponseUtil.returnBadRequest(
-                        res,
-                        `Class information could not be found`,
-                    );
-                    return;
-                }
-
-                const mailerPromises = [];
-                waitlistRecords.forEach((record) => {
-                    const emailSubject = `Spot Available for Waitlisted Class: ${waitlistClass.name}`;
-                    mailerPromises.push(
-                        send(
+            if (notifyWaitlist && waitlistRecords.length > 0) {
+                await Promise.all(
+                    waitlistRecords.map((record) => {
+                        const emailSubject = `Spot Available for Waitlisted Class: ${waitlistClass.name}`;
+                        return send(
                             process.env.EMAIL_FROM,
                             (record as any).parent.user.email,
                             emailSubject,
@@ -211,12 +216,11 @@ export default async function handle(
                                 waitlistClass.endDate,
                                 waitlistClass.startTimeMinutes,
                                 waitlistClass.durationMinutes,
-                                (record as any).parent.preferredLanguage,
+                                record.parent.preferredLanguage,
                             ),
-                        ),
-                    );
-                });
-                await Promise.all(mailerPromises);
+                        );
+                    }),
+                );
             }
 
             ResponseUtil.returnOK(res, deletedRegistration);
--- a/pages/api/enroll/child.ts
+++ b/pages/api/enroll/child.ts
@@ -15,7 +15,6 @@ import { getClass } from "@database/class";
 import { getWaitlistRecordsByClassId } from "@database/waitlist";
 import send from "services/nodemailer/mail";
 import { openSpotWaitlistTemplate } from "@utils/mail/templateWaitlist";
-import { IoRecordingOutline } from "react-icons/io5";
 
 /**
  * handle controls the request made to the enroll/child resource.
@@ -167,23 +166,6 @@ export default async function handle(
                 return;
             }
 
-            // check if the class is full, prior to deletion
-            let notifyWaitlist = true;
-            const waitlistClass = await getClass(
-                parentRegistrationInput.classId,
-            );
-            if (!waitlistClass) {
-                ResponseUtil.returnBadRequest(
-                    res,
-                    `Class information could not be found`,
-                );
-                return;
-            }
-            // if class wasn't originally full, do not notify waitlist
-            if (waitlistClass._count.parentRegs < waitlistClass.spaceTotal) {
-                notifyWaitlist = false;
-            }
-
             const deletedRegistration = await deleteParentRegistration(
                 parentRegistrationInput,
             );
@@ -195,6 +177,15 @@ export default async function handle(
                 return;
             }
 
+            let notifyWaitlist = true;
+            // if class wasn't originally full, do not notify waitlist
+            if (
+                deletedRegistration.class._count.parentRegs <
+                deletedRegistration.class.spaceTotal
+            ) {
+                notifyWaitlist = false;
+            }
+
             // check if there exists waitlist records for this class
             const waitlistRecords = await getWaitlistRecordsByClassId(
                 parentRegistrationInput.classId,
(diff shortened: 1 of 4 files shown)


pages/api/enroll/child.ts
```diff
@@ -174,6 +177,48 @@ export default async function handle(
...
+                                waitlistClass.startTimeMinutes,
+                                waitlistClass.durationMinutes,
+                                (record as any).parent.preferredLanguage,
+                            ),
+                        ),
+                    );
+                });
+                await Promise.all(mailerPromises);
+            }
+
```
aamyli, line 217: i can extract this into its own function under `utils` or something if we want, just not sure if necessary

utils/mail/templateWaitlist.ts
```diff
@@ -0,0 +1,105 @@
...
+                <p>Use this link to remove yourself from the class waitlist:</p>
+                <a href="${process.env.NEXTAUTH_URL}/class">${
+        process.env.NEXTAUTH_URL
+    }/class</a>
+                <br></br>
+                <br></br>
+                <p>Regards, Social Diversity for Children</p>
+            </body>
+            `;
+};
```
aamyli, line 109: Screen Shot 2021-11-14 at 7 29 16 PM

for reference, this is what the current email looks like. i'm not sure if this was the best way to word everything lol let me know your thoughts!
@bekahma @vedantdsn @MatoPlus

aamyli: something to note is that this email is sent _**every time**_ an unenrollment occurs -- that means if eg a parent unenrolls 3 children at once, everyone will get 3 emails ...

rickson and i were talking about just wording it in a way so that it doesn't seem spammy and is like... indicating that a spot has opened up ?? so 3 emails indicates 3 spots have opened up ?? could this be worded better to align with that / should we think about implementing something to prevent multiple emails within a short time range?

pages/api/enroll/child.ts
```diff
@@ -174,6 +177,48 @@ export default async function handle(
...
+                    return;
+                }
+
+                const mailerPromises = [];
+                waitlistRecords.forEach((record) => {
+                    const emailSubject = `Spot Available for Waitlisted Class: ${waitlistClass.name}`;
+                    mailerPromises.push(
+                        send(
+                            process.env.EMAIL_FROM,
+                            (record as any).parent.user.email,
```
aamyli: oh yeah also a point about this ... i was getting an error saying "property does not exist on value of type" and was reading about how to fix ([stackoverflow](https://stackoverflow.com/questions/18083389/ignore-typescript-errors-property-does-not-exist-on-value-of-type)) --

if i leave it as `record.parent.user.email` i just get the red squiggly line under `parent` but everything executes fine; i'll take a closer look at this error later + get it fixed before merge !

utils/mail/templateWaitlist.ts
MatoPlus: Yea curious as to what the designers think. I feel like we could get away by changing the main header in the mail to `A spot opened up for <Class>!`

pages/api/enroll/child.ts
MatoPlus: yea this is an instance of prisma not playing well with typescript. You can get away with this by telling typescript to trust Prisma return types. 

```typescript
// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
async function getWaitlistRecordsByClassId(classId: number) {
```

then, the type returned will have `.parent` as you expect.

pages/api/enroll/child.ts
MatoPlus: yeaa code does look a bit chunky...

If you have something in mind, go for it!

utils/mail/templateWaitlist.ts
aamyli: i was thinking of that phrase but i wasn't sure if there'd be any ambiguity on what it meant (is "opened up" as easily understandable to non-native speakers?), and there seems to be no intent on translating ? (considering our target audience + languages)

MatoPlus: That's a very good point. We could ask for translations actually (especially if the designers are set on something), if not, I think it's fine to leave it as you worded as well

pages/api/enroll/child.ts
```diff
@@ -174,6 +177,48 @@ export default async function handle(
...
+                                waitlistClass.startDate,
+                                waitlistClass.endDate,
+                                waitlistClass.startTimeMinutes,
+                                waitlistClass.durationMinutes,
+                                (record as any).parent.preferredLanguage,
+                            ),
+                        ),
+                    );
+                });
+                await Promise.all(mailerPromises);
```
MatoPlus: I wonder if you could just wrap this into 1 call via map
```suggestion
                await Promise.all(
                waitlistRecords.map((record) => {
                    const emailSubject = `Spot Available for Waitlisted Class: ${waitlistClass.name}`;
                    return(
                        send(
                            process.env.EMAIL_FROM,
                            (record as any).parent.user.email,
                            emailSubject,
                            openSpotWaitlistTemplate(
                                req.body.classId,
                                waitlistClass.imageLink,
                                waitlistClass.name,
                                waitlistClass.weekday,
                                waitlistClass.startDate,
                                waitlistClass.endDate,
                                waitlistClass.startTimeMinutes,
                                waitlistClass.durationMinutes,
                                (record as any).parent.preferredLanguage,
                            ),
                        ),
                    );
                }););
```

^^ something like that.

pages/api/enroll/child.ts
```diff
@@ -174,6 +177,48 @@ export default async function handle(
...
+                            openSpotWaitlistTemplate(
+                                req.body.classId,
+                                waitlistClass.imageLink,
+                                waitlistClass.name,
+                                waitlistClass.weekday,
+                                waitlistClass.startDate,
+                                waitlistClass.endDate,
+                                waitlistClass.startTimeMinutes,
+                                waitlistClass.durationMinutes,
+                                (record as any).parent.preferredLanguage,
```
MatoPlus: idem

services/database/waitlist.ts
```diff
@@ -30,6 +30,13 @@ async function getWaitlistRecordsByClassId(
         where: {
```
MatoPlus, line 29: please make the change I suggested in my previous comment for getting around types.

MatoPlus (changes requested):
Just a few suggestions and I think this is good to merge!

bekahma (commented):
email template looks good to me!

we def don't want to send out an email every time someone removes themselves from a class. maybe only send out a 2nd email if the class is available AFTER it's full again (the logic might be wonky here though)? also in the email template maybe we can mention that just because a spot opened up doesn't mean it's guaranteed? ex. if a parent waited 2 hours to sign up but someone already took the spot

aamyli:
Screen Shot 2021-11-16 at 5 45 10 PM

added logic so that the email will only be sent if an unenrollment occurs from a FULL class -> ie if the class was not full before, no email will be sent @MatoPlus ; i tested this out

also added a comment in the email that says registration is first come first serve, lmn if you have different wording in mind to signal that registration is not guaranteed @bekahma 

i am good to merge this PR if both of you approve!

bekahma:
yes thanks for making the edit amy!!

pages/api/enroll/child.ts
```diff
@@ -12,6 +12,10 @@ import { ParentRegistrationInput } from "models/Enroll";
 import { roles } from "@prisma/client";
 import { getUserFromEmail } from "@database/user";
 import { getClass } from "@database/class";
+import { getWaitlistRecordsByClassId } from "@database/waitlist";
+import send from "services/nodemailer/mail";
+import { openSpotWaitlistTemplate } from "@utils/mail/templateWaitlist";
+import { IoRecordingOutline } from "react-icons/io5";
```
MatoPlus: @aamyli Please try to remove github action warning when ever possible. It doesn't look like this icon is being used.

pages/api/enroll/child.ts
```diff
@@ -163,6 +167,23 @@ export default async function handle(
...
+                return;
+            }
+            // if class wasn't originally full, do not notify waitlist
+            if (waitlistClass._count.parentRegs < waitlistClass.spaceTotal) {
+                notifyWaitlist = false;
+            }
+
             const deletedRegistration = await deleteParentRegistration(
                 parentRegistrationInput,
             );
```
MatoPlus: This works but I think we can squish this into one transaction by returning the count in the `deleteParentRegistration` call and tweak the logic to say if the reg count is total space - 1, then we notify.

Kinda optional but it would speed up our enrollments a bit 😄 

@aamyli  thoughts?

MatoPlus (approved):
Looks great! Thanks for getting this ticket in @aamyli 🔥 

Just got a few more optional suggestions. If you think you've tested things thoroughly, feel free to merge this in!

PR #139, opened 2021-11-19
Mg/terraform

matthewgeng:
### Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/6dd603d3687743ce8d09d6298dc99748?v=7332b6570bcc4b48b5cd73fa2719bec8&p=a7e4ef1d87a5455da91b0de58d09bd14)

* update gitignore
    * gitignore all files inside lambda with the exception of all .py files
    * terraform files that aren't needed
* renamed api mailing lambda key from AWS_LAMBDA_KEY to LAMBDA_KEY_SECRET because SSM doesn't allow variables prefixed with SSM or AWS (case insensitive)
* added all terrraform files 
* terraform structure
    * environments
        * dev
            * remote_state/
        * prod
    * modules
        * eventbridge
        * iam
        * lambda
        * parameter_store
        * s3
*  remote state (must run this before the main terraform files to setup the backend, famous terraform chicken/egg problem)
* currently we grouped modules by service but this should probably change to higher levels of abstraction since essentially all the modules are being defined in main.tf which could become cluttered
* reminder anything defined or even retrieved like with the `data` keyword is stored in terraform state in plaintext
    * LAMBDA_SECRET_KEY is retrieved (only kind of sensitive value right now) not defined (this is the only thing that must be created prior to running terraform since we don't want to define it)
    * We must treat the terraform state as a secret itself
    

### Steps to test

1. apply remote state backend (cd remote_state && terraform apply)
2. create LAMBDA_SECRET_KEY in parameter store
3. terraform apply in dev environment

* note infrastructure by nature is hard to test so testing might not actually be possible

-

### Checklist

-   [x] I have run the linter


.gitignore
```diff
@@ -12,4 +13,18 @@ node_modules
...
+
+# logs
+*.log
+
+# Directories
+.terraform/
+
+# lambda
+**/cronMailing/*
+!**/cronMailing/*.py
```
MatoPlus: can you help me add a new line here, assuming eslint doesn't complain
```suggestion
!**/cronMailing/*.py

```

terraform/environments/dev/.terraform.lock.hcl
```diff
@@ -0,0 +1,38 @@
+# This file is maintained automatically by "terraform init".
```
MatoPlus, line 1: do we need this file?

terraform/environments/dev/remote_state/.terraform.lock.hcl
```diff
@@ -0,0 +1,20 @@
+# This file is maintained automatically by "terraform init".
```
MatoPlus, line 1: do we need this file?

terraform/environments/dev/remote_state/main.tf
```diff
@@ -0,0 +1,55 @@
...
+
+  attribute {
+    name = "LockID"
+    type = "S"
+  }
+
+  tags = {
+    name = "Terraform Remote Dynamo State Lock"
+  }
+}
```
MatoPlus: new line plz

terraform/environments/dev/terraform.tfvars
```diff
@@ -0,0 +1,30 @@
...
+zoom_class_link_name        = "zoom-class-link"
+zoom_class_link_value       = "https://us06web.zoom.us/j/8391781258?pwd=NHBlMVh5SHc5UE10K2dsMFFvTzdmUT09"
+zoom_class_link_type        = "String"
+
+# ------------------------------------------------------------------
+# Lambda
+lambda_runtime = "python3.8"
+# cronMailing
+cronMailing_rule_name           = "hourly-cron-mailing"
+cronMailing_schedule_expression = "cron(0 * ? * * *)"
```
MatoPlus: is this tested? isn't this supposed to be:
```suggestion
cronMailing_schedule_expression = "cron(0 * * * *)"
```

terraform/modules/eventbridge/main.tf
```diff
@@ -0,0 +1,12 @@
+
+
```
MatoPlus:
```suggestion
```
removing empty lines

MatoPlus (commented):
If possible can you document this folder with a README.md
I think it would be really helpful for future references!

MatoPlus (commented):
I haven't looked at tf file contents closely but I'll definitely do it later tomorrow or sat and let you know!

terraform/environments/dev/.terraform.lock.hcl
matthewgeng: Yeah you can think of it as package.json, it essentially makes sure every time you do a terraform init you have the same provider version

terraform/environments/dev/remote_state/.terraform.lock.hcl
matthewgeng: ^

terraform/environments/dev/.terraform.lock.hcl
MatoPlus: ahh makes sense, ty.

It's just kinda funny that my experience in startups, I usually don't see this file around in self-managed files.

I feel like it wouldn't hurt to include it though.

terraform/environments/dev/remote_state/main.tf
```diff
@@ -0,0 +1,55 @@
...
+    rule {
+      apply_server_side_encryption_by_default {
+        sse_algorithm = "AES256"
+      }
+    }
+  }
+}
+
+# State locks for terraform with dynamo
+resource "aws_dynamodb_table" "sdc_terraform_state_lock" {
```
MatoPlus, line 39: I don't think dynamodb has a forever free tier, but I guess it is pretty cheap 👀

terraform/environments/dev/main.tf
```diff
@@ -0,0 +1,72 @@
...
+  # Parameter store (ssm) inputs
+  # zoom link
+  zoom_class_link_description = var.zoom_class_link_description
+  zoom_class_link_name        = var.zoom_class_link_name
+  zoom_class_link_value       = var.zoom_class_link_value
+  zoom_class_link_type        = var.zoom_class_link_type
+
+  # lambda secret 
+  lambda_secret_key_name = var.lambda_secret_key_name
+}
```
MatoPlus, line 54: Up to you but I feel like we could separate out each of these modules in their own files like `dev/ssm.tf` and in the main file, we can include more generic/general modules like iam, generic buckets, providers, secrets.

terraform/environments/dev/variables.tf
```diff
@@ -0,0 +1,100 @@
...
+
+variable "cronMailing_rule_name" {
+  description = "Name of cronMailing eventbridge rule"
+  default     = "hourly-cron-mailing"
+  type        = string
+}
+
+variable "cronMailing_schedule_expression" {
+  description = "CRON or Rate expression for cronMailing job"
+  default     = "cron(0 * ? * * *)" # hourly cron expression
```
MatoPlus, line 108: idem

terraform/environments/dev/terraform.tfvars
```diff
@@ -0,0 +1,30 @@
+# ------------------------------------------------------------------
```
MatoPlus: hmm I might not be catching something but isn't this file redundant if we already stated var defaults in the `variables.tf` file?

I think we usually want to ignore all tfvars as they include sensitive data. Lambda secret would be a great example of this case.

I feel like we can keep these files in some github secret and only allow ci to access and input them securely. But that might be a bit expensive given our time. I feel like we can just keep an internal terraform.tfvars in some kind of vault (or you can just pass it on to me for safe keepings).

.gitignore
```diff
@@ -12,4 +13,18 @@ node_modules
...
 # DS _store
 .DS_Store
-.vscode
\ No newline at end of file
+
+# unnecessary terraform files
+*.tfstate
+*.tfstate.backup
+*.tfstate.lock.info
+
```
MatoPlus, line 22: from github gitignore samples

```suggestion

# Exclude all .tfvars files, which are likely to contain sentitive data, such as
# password, private keys, and other secrets. These should not be part of version
# control as they are data points which are potentially sensitive and subject
# to change depending on the environment.
#
*.tfvars
```

terraform/environments/dev/variables.tf
```diff
@@ -0,0 +1,100 @@
...
+  type        = string
+}
+
+# ------------------------------------------------------------------
+# environment variables
+
+variable "api_endpoint" {
+  description = "Domain api endpoint"
```
MatoPlus, line 116: I feel like it's fine to add the default here for the mail route since our code is open source, thoughts?

terraform/modules/eventbridge/main.tf
```diff
@@ -0,0 +1,12 @@
...
+resource "aws_cloudwatch_event_rule" "this" {
```
MatoPlus: do we have to name the resource as "this"? would a more explicit noun like "rule" be better?
```suggestion
resource "aws_cloudwatch_event_rule" "rule" {
```

terraform/modules/eventbridge/main.tf
```diff
@@ -0,0 +1,12 @@
...
+  name                = var.rule_name
+  schedule_expression = var.schedule_expression
+}
+
+resource "aws_cloudwatch_event_target" "this" {
```
MatoPlus: idem

terraform/modules/lambda/files/cronMailing/cronMailing.py
```diff
@@ -0,0 +1,13 @@
+import json
+import os
+import requests
+
+def lambda_handler(event, context):
+    print("*** START MAILING CRON JOB ***")
+    
+    res = requests.post(os.environ.get("API_ENDPOINT"), data = { "key": os.environ.get("LAMBDA_SECRET_KEY") })
+    print(res.text)
+
```
MatoPlus: @matthewgeng this was tested right?

terraform/modules/lambda/lambda.tf
```diff
@@ -0,0 +1,25 @@
+data "archive_file" "lambda_zip" {
+  type        = "zip"
+  source_dir  = "${path.module}/files/${var.function_name}"
+  output_path = "${path.module}/files/zips/${var.function_name}.zip"
+}
+
+resource "aws_lambda_function" "this" {
```
MatoPlus, line 7: idem

terraform/modules/s3/s3.tf
```diff
@@ -0,0 +1,71 @@
...
+resource "aws_s3_bucket" "s3_uploads" {
+  bucket = var.s3_bucket_name
+  # TODO look into better policies later
+  acl = "private"
+
+  # TODO improve this security wise
+  cors_rule {
+    allowed_headers = ["*"]
+    allowed_methods = ["PUT", "POST", "DELETE"]
+    allowed_origins = ["http://localhost:3000", "https://staging.socialdiversity.org"]
```
MatoPlus: can we also allow all pr deploys here? Did you run into any issues?

MatoPlus (changes requested):
Some suggestions, great work overall @matthewgeng! I'll take another look sometime this weekend.

terraform/environments/dev/terraform.tfvars
matthewgeng: mm yeah I tested the expression and it showed me the correct times, but yeah I'll change it!. It was from an amazon doc on sagemaker which I don't think completely applies https://docs.aws.amazon.com/sagemaker/latest/dg/model-monitor-schedule-expression.html

terraform/environments/dev/remote_state/main.tf
matthewgeng: ah didn't know about that. I'm not sure there is a workaround? This is what everybody does for terraform :(

terraform/modules/lambda/files/cronMailing/cronMailing.py
matthewgeng: yes

terraform/modules/eventbridge/main.tf
matthewgeng: this name is for local reference only and terraform suggests a naming convention that doesn't repeat information from the resource type https://www.terraform-best-practices.com/naming. However I do agree I don't completely like "this". This pattern is widely used in popular terraform modules

terraform/modules/s3/s3.tf
matthewgeng: I'm not sure we can add pr deploys here since the pr domains would be dynamic right? And would probably want to be added and removed in the pr lifecycle? I'm actually not sure what to do about this

terraform/modules/eventbridge/main.tf
matthewgeng: ^

terraform/modules/lambda/lambda.tf
matthewgeng: ^

terraform/environments/dev/variables.tf
matthewgeng: Not sure honestly. I'm fine with both

terraform/environments/dev/remote_state/main.tf
MatoPlus: meh it's probably fine. I think the usage should just be 1-3 dollars/month at worst hopefullly. Do keep an eye on the billing though 😉

terraform/environments/dev/variables.tf
MatoPlus: Let's add it as a default. It's defined per environment.

terraform/modules/s3/s3.tf
MatoPlus: > Access-Control-Allow-Origin Headers can only be a wildcard '*' or an exact url. 

hmm, okay. Let's just keep PR deploys out of this.

matthewgeng: Oh true the wildcard! I think that’s possible

terraform/environments/dev/variables.tf
matthewgeng: Sound good

terraform/modules/eventbridge/main.tf
MatoPlus: hmm, I see. In that case, I think we can just leave it, thanks for the src!

terraform/environments/dev/terraform.tfvars
```diff
@@ -0,0 +1,30 @@
...
+criminal_check_folder   = "criminal-check"
+income_proof_folder     = "income-proof"
+curriculum_plans_folder = "curriculum-plans"
+other_folder            = "other"
+# ------------------------------------------------------------------
+# Parameter Store (SSM)
+# zoom-class-link
+zoom_class_link_description = "Link used to access virtual classes hosted over Zoom"
+zoom_class_link_name        = "zoom-class-link"
+zoom_class_link_value       = "https://us06web.zoom.us/j/8391781258?pwd=NHBlMVh5SHc5UE10K2dsMFFvTzdmUT09"
```
MatoPlus: leave out for now

PR #144, opened 2021-11-26
Add all dev terraform infra

matthewgeng:
### Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/6dd603d3687743ce8d09d6298dc99748?v=7332b6570bcc4b48b5cd73fa2719bec8&p=a7e4ef1d87a5455da91b0de58d09bd14)

* update gitignore
    * gitignore all files inside lambda with the exception of all .py files
    * terraform files that aren't needed
* renamed api mailing lambda key from AWS_LAMBDA_KEY to LAMBDA_KEY_SECRET because SSM doesn't allow variables prefixed with SSM or AWS (case insensitive)
* added all terrraform files 
* terraform structure
    * environments
        * dev
            * remote_state/
        * prod
    * modules
        * eventbridge
        * iam
        * lambda
        * parameter_store
        * s3
*  remote state (must run this before the main terraform files to setup the backend, famous terraform chicken/egg problem)
* currently we grouped modules by service but this should probably change to higher levels of abstraction since essentially all the modules are being defined in main.tf which could become cluttered
* reminder anything defined or even retrieved like with the `data` keyword is stored in terraform state in plaintext
    * LAMBDA_SECRET_KEY is retrieved (only kind of sensitive value right now) not defined (this is the only thing that must be created prior to running terraform since we don't want to define it)
    * We must treat the terraform state as a secret itself
    

### Steps to test

1. apply remote state backend (cd remote_state && terraform apply)
2. create LAMBDA_SECRET_KEY in parameter store
3. create zoom-class-link in parameter store
4. add secret `terraform.tfvars` (you'll get from us) file into `terraform/environments/dev` (location where main.tf is)
5. terraform apply in dev environment

* note infrastructure by nature is hard to test so testing might not actually be possible

-

### Checklist

-   [x] I have run the linter

### DIFF
--- a/terraform/environments/dev/variables.tf
+++ b/terraform/environments/dev/variables.tf
@@ -1,0 +1,124 @@
+# NOTE
+# Variables without a default are inputted secretly. Generally only define variables with a default for non-secrets
+
+# ------------------------------------------------------------------
+# app
+variable "sdc_domain" {
+  description = "full domain name"
+  default     = "https://staging.socialdiversity.org"
+  type        = string
+}
+
+variable "sdc_pr_domain" {
+  description = "domain of pr deploys"
+  default     = "https://social-diversity-for-children-pr-*.up.railway.app"
+  type        = string
+}
+
+# should add variable condition checks to limit configuration discrepencies
+# ------------------------------------------------------------------
+# IAM
+variable "cloudwatch_lambda_logs_policy_name" {
+  description = "Name of cloudwatch lambda logs policy"
+  default     = "cloudwatch-lambda-logs-policy"
+  type        = string
+}
+
+# ------------------------------------------------------------------
+# s3
+
+variable "s3_images_bucket_name" {
+  description = "Name of the document uploads s3 bucket"
+  default     = "sdc-public-images"
+  type        = string
+}
+
+variable "s3_uploads_bucket_name" {
+  description = "Name of the document uploads s3 bucket"
+  default     = "sdc-uploads"
+  type        = string
+}
+
+variable "criminal_check_folder" {
+  description = "Name of the criminal check folder"
+  default     = "criminal-check"
+  type        = string
+}
+
+variable "income_proof_folder" {
+  description = "Name of the income proof folder"
+  default     = "income-proof"
+  type        = string
+}
+
+variable "curriculum_plans_folder" {
+  description = "Name of the curriculum plans folder"
+  default     = "curriculum-plans"
+  type        = string
+}
+
+variable "other_folder" {
+  description = "Name of the other folder (folder where files are uploaded when type isn't specified)"
+  default     = "other"
+  type        = string
+}
+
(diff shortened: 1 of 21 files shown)


terraform/environments/dev/variables.tf
```diff
@@ -0,0 +1,124 @@
...
+  default     = "cron(0 * ? * * *)" # hourly cron expression
+  type        = string
+}
+
+# ------------------------------------------------------------------
+# environment variables
+
+variable "api_endpoint" {
+  description = "Domain api endpoint"
+  default     = "https://staging.socialdiversity.org/api"
```
MatoPlus, line 117: isn't this suppose to point to the mail api endpoint?
```suggestion
  default     = "https://staging.socialdiversity.org/api/mail"
```

terraform/modules/lambda/files/cronMailing/cronMailing.py
```diff
@@ -0,0 +1,12 @@
...
+import requests
+
+def lambda_handler(event, context):
+    print("*** START MAILING CRON JOB ***")
+    
+    res = requests.post(f'{os.environ.get("API_ENDPOINT")}/mail', data = { "key": os.environ.get("LAMBDA_SECRET_KEY") })
+    print(res.text)
+
+    print("*** END MAILING CRON JOB ***")
+
```
MatoPlus: add newline for eof plz

MatoPlus (changes requested):
Got a couple of nits.

I have not looked too deeply into the terraform files but this looks pretty good to me overall. @matthewgeng is there anything you are concerned with atm? Otherwise we can merge this and potentially blitz for infra bugs in the incoming days before the final MVP demo next week.

terraform/environments/dev/variables.tf
matthewgeng: I changed it to a general api and appended ‘mail’ in the lambda itself since I thought that was better. But I change it back if you want

terraform/modules/lambda/files/cronMailing/cronMailing.py
matthewgeng: isn't there a newline already?

terraform/environments/dev/variables.tf
MatoPlus: ah okay, that's fine then. no revert needed

terraform/modules/lambda/files/cronMailing/cronMailing.py
MatoPlus: I thought so as well 👀 
image

MatoPlus:
```suggestion
    print("*** END MAILING CRON JOB ***")

```

MatoPlus (approved):
LGTM!

PR #145, opened 2021-11-26
feature - parent registration filter out non eligible students

aamyli:
### Notion ticket link

[parent registration - filter out non-eligible participants](https://www.notion.so/uwblueprintexecs/Parent-registration-Filter-out-not-eligible-participants-2c1267bbd5a04935a9f0a7e3a2c2e660)

[loom demo](https://www.loom.com/share/8a0ee978524b4a97bfd668339ed3160a)

Screen Shot 2021-11-25 at 11 00 18 PM

- Added check to see whether each child is eligible age-wise for the specified program

### Steps to test

1. Can alter the ages and see if the button gets disabled accordingly

- refer to [comment](https://github.com/uwblueprint/social-diversity-for-children/pull/145/files#r757222408)

### Checklist

-   [x] I have run the linter


src/components/SelectChildForClass.tsx
```diff
@@ -28,6 +36,7 @@ export default function SelectChildForClass(
                     </b>
```
aamyli: ^^ i marked this area bc currently the title of the class is hardcoded, i could pass in the class name as a param and fix it in this ticket if we want (not sure if we have an existing ticket for it, if we do i couldn't find it)

aamyli: lines 34-35^

MatoPlus: Yea good catch! Please patch this up for the team 🙏

src/components/SelectChildForClass.tsx
```diff
@@ -12,6 +13,12 @@ type SelectChildForClassProps = {
 export default function SelectChildForClass(
     props: SelectChildForClassProps,
 ): JSX.Element {
+    if (!props.eligible[props.selectedChild]) {
+        let index = 0;
+        while (!props.eligible[index] && index < props.eligible.length) index++;
+        props.setSelectedChild(index);
+    }
+
```
aamyli, line 23: so i needed to add a check so that the default `selectedChild` on page load was actually an eligible child -- this works UNLESS no children are eligible for a particular class, in which then the code breaks (assumingly because the index goes out of bounds) 

the way the current flow is set up (as the `selectedChild` is passed onto the next page), i'm not sure if there's a way around it... another option could to be if all children are ineligible, we direct to a different page saying that none of their children are eligible for this program?

MatoPlus: doing a `router.back()` + error toast message makes sense to me!

aamyli: i just realized that if no children are eligible, then the program won't be clickable in the first place  (see below), so do we still need a router call? The only time where there'd be an issue is if the user tried to directly go to the link (http://localhost:3000/parent/enrollment?classId=10001) instead of clicking to it -- 

Screen Shot 2021-11-26 at 12 24 30 AM

I tried doing a router thing anyways to see if it'd work but it's giving me an error -- when we call a router.back() does it break away from the rest of the code? Because even with a router.back() call to home, I still get an error in the `ClassEnrollmentConfirm` page.

MatoPlus: @aamyli  Yes, we always want to consider all edge cases like you described. Using router takes some time to load (async), so what I usually do is in the clause where we are routing to somewhere, I also return a <CommonLoading /> like:
```typescript
router.back();
return <CommonLoading />
```

to avoid reaching the "bad" that we are trying to avoid by using router.back().

src/components/SelectChildForClass.tsx
```diff
@@ -20,12 +28,8 @@ export default function SelectChildForClass(
...
             </Center>
             <Center>
-                <Text pb="55px" align="center" mt="30px">
-                    Who would you like to register for{" "}
-                    <b>
-                        Building Bridges with Music - <br />
-                        Singing Monkeys (Ages 9 and under)?{" "}
-                    </b>
+                <Text pb="55px" align="center" mt="30px" width="60%">
+                    Who would you like to register for <b>{props.className}</b>
```
MatoPlus, line 33: I think you also need the program name here. Also, please rebase and use i18n for translation integration:
```suggestion
                    {t("enroll.selectChild")} <b>{props.programName} - {props.className} ({t(
                                    props.isAgeMinimal
                                        ? "program.ageGroupAbove"
                                        : "program.ageGroupUnder",
                                    {
                                        age: props.borderAge,
                                    },
                                )})</b>
```

we can ignore the `(select one)` for now. 

^ where t is the object in the useTranslation hook for the form translations (should be in the file once you rebase)

NOTE: Since we are using quite a lot of props from class, we might just want to pass the class itself as a prop into this component.

src/components/SelectChildForClass.tsx
aamyli: there's no point of a toast then, right? because it won't always show depending on what it's routing back to (eg. i went to the URL directly from a google page so no toast showed up)

src/components/SelectChildForClass.tsx
aamyli: @MatoPlus i already did that in the other file [here](https://github.com/uwblueprint/social-diversity-for-children/pull/145/files#diff-83ff1da87167756ae59b558afa4424bea36de0102fedd2a29dc2a725e93c1608R113) - lines 113-121 ; will rebase for the `enroll.selectChild` part

src/components/SelectChildForClass.tsx
MatoPlus: mm, yea that's true. Let's leave the toast out!

MatoPlus:
equiv merged with https://github.com/uwblueprint/social-diversity-for-children/commit/7fd51c81004b438daf02144b018a287d505e3bff

PR #157, opened 2021-11-29
Terraform ci with files changed composite action

matthewgeng:
### Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/Implement-CI-CD-pipeline-for-terraform-f6ed9285f2f646b18fcb9296bb9253a5)

* We call a composite action (closest to a published action) for checking if there is a modified file that matches a regular expression
    * check-files-changed action is defined in another [repo](https://github.com/uwblueprint/sdc-actions/blob/main/actions/check-files-changed/action.yml) (how it should be done)
    * FILES DELETED ARE NOT DETECTED AS A CHANGE, this is set in the other repo but could change to an input with a default if functionality is needed
    * this action can be used by other teams or technically anyone in the world since it's public
* didn't do OIDC stuff yet (something to look into for future, made a [ticket](https://www.notion.so/uwblueprintexecs/6dd603d3687743ce8d09d6298dc99748?v=7332b6570bcc4b48b5cd73fa2719bec8&p=d27463c309144f27bcd9c92152c8022e))
* Believe there are some redundancies in the comment creation (hashicorp's docs on the terraform ci) the comment show's status of terraform format, init, and validate however if they failed we would never get to the part where the comment is created so kind of redundant imo (maybe we should not have? idk)
* In the the hashicorp docs they essentially have the same pipeline for pushes (they simply check at the end if it's a push and do a terraform apply). Since I've split up PR's and pushes, there would be some code duplication for terraform at least. Maybe we could go back to the single ci file? Not sure
* Currently we only check if files need to be formatted, we don't do any formatting in the PR, maybe we should do formatting or just on a push?
* Added a staging environment in github with secrets
    * AWS_ACCESS_KEY_ID
    * AWS_REGION
    * AWS_SECRET_ACCESS_KEY
    * LAMBDA_SECRET_KEY_NAME
    * ZOOM_CLASS_LINK_VALUE
* Added production environment with draft secrets
    * AWS_REGION
* ON PUSH:
    * we only do a terraform init, plan, and apply (NO FMT OR VALIDATE)

### Steps to test

1.  Make changes to terraform and make a pr/push

-

### Checklist

-   [x] I have run the linter


.github/workflows/ci-pr.yml
```diff
@@ -1,73 +1,185 @@
...
+            echo "::set-output name=env::staging"
+            echo Environment set to 'staging'
+          elif [[ $GITHUB_BASE_REF = "main" ]] 
+          then
+            echo "::set-output name=env::production"
+            echo Environment set to 'production'
+          else:
+            echo '$GITHUB_BASE_REF' does not equal 'staging' or 'main'
+            exit 1
+          fi
```
MatoPlus: this is shell script right? I would encourage running `shellfmt` and `shellcheck` whenever you're dealing with shell : )

```suggestion
          if [[ $GITHUB_BASE_REF = "staging" ]]; then
              echo "::set-output name=env::staging"
              echo Environment set to 'staging'
          elif [[ $GITHUB_BASE_REF = "main" ]]; then
              echo "::set-output name=env::production"
              echo Environment set to 'production'
          else
              echo "'$GITHUB_BASE_REF' does not equal 'staging' or 'main'"
              exit 1
          fi
```

.github/workflows/ci-pr.yml
```diff
@@ -1,73 +1,185 @@
...
+
+      # TODO change to OIDC provider
+      - name: Configure AWS Credentials 
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.AWS_REGION }}
+      
+      # TODO should we format as well? Not just check?
```
MatoPlus: this is fine.
```suggestion
```

.github/workflows/ci-push.yml
```diff
@@ -1,50 +1,156 @@
...
+            echo "::set-output name=env::staging"
+            echo Environment set to 'staging'
+          elif [[ $GITHUB_BASE_REF = "main" ]] 
+          then
+            echo "::set-output name=env::production"
+            echo Environment set to 'production'
+          else:
+            echo '$GITHUB_BASE_REF' does not equal 'staging' or 'main'
+            exit 1
+          fi
```
MatoPlus: shellcheck fix!
```suggestion
          if [[ $GITHUB_BASE_REF = "staging" ]]; then
              echo "::set-output name=env::staging"
              echo Environment set to 'staging'
          elif [[ $GITHUB_BASE_REF = "main" ]]; then
              echo "::set-output name=env::production"
              echo Environment set to 'production'
          else
              echo "'$GITHUB_BASE_REF' does not equal 'staging' or 'main'"
              exit 1
          fi
```

.github/workflows/ci-pr.yml
```diff
@@ -1,73 +1,185 @@
...
+          terraform plan -no-color
+        continue-on-error: true
+
+      - name: Update Pull Request
+        uses: actions/github-script@0.9.0
+        env:
+          PLAN: "terraform\n${{ steps.plan.outputs.stdout }}"
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          # not sure if there is much point in having fmt, init and validate output here since if they failed we would never get to this point
```
MatoPlus: I would argue it's fine for assurance.
```suggestion
```

.github/workflows/ci-push.yml
```diff
@@ -1,50 +1,156 @@
...
+          terraform apply -auto-approve tfplan
+        continue-on-error: true
+
+      - name: Update Pull Request
+        uses: actions/github-script@0.9.0
+        env:
+          PLAN: "terraform\n${{ steps.plan.outputs.stdout }}"
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          # not sure if there is much point in having fmt, init and validate output here since if they failed we would never get to this point
```
MatoPlus: I would argue it's fine for assurance.
```suggestion
```

MatoPlus (approved):
Looking good, no major suggestions! Great work 🔥

PR #173, opened 2021-12-07
Add confirm alert dialog + Loading Prompt For Unregistering class

neelismail01:
### Notion ticket link

[Add confirm alert dialog + Loading Prompt For Unregistering class](https://www.notion.so/uwblueprintexecs/Add-confirm-alert-dialog-Loading-Prompt-For-Unregistering-class-1e8abb7901084dc081dd046e19f3dc76)

In this ticket, I was responsible for creating a confirmation Modal, so that users could verify that they wanted to unregister a child from a class before doing so. I achieved this by customizing a Modal from the Chakra UI library, which would get opened whenever a user clicked on a child to unregister or the "Unregister All" button (See video below for demonstration of user interaction). The student(s) to be unregistered would be stored in the component state variable called "studentsToUnregister" while the user confirms that they want to unregister a student. If the user confirms the unregistration, the student would be passed to a function deleteClassRegistration() or deleteClassRegistrations() - which had already been implemented - depending if a single or multiple students were selected to be unregistered. After the user confirms their decision to unregister a student, the component variable storing the children to unregister is restored to its default state.

### Steps to test

https://user-images.githubusercontent.com/23081661/145119797-5cd03ea1-eade-4ee7-bf48-e6bd79fa1665.mov

1. Validate that when a user unregisters a single child, only that child is unregistered
2. Validate that when a user unregisters all children, all of the children are unregistered
3. Validate that when a user clicks "No" on the confirmation modal, none of the children are unregistered from the program 

- DESIGNERS: do you have any comments/recommendations for the design of the modal or does everything look good?

### Checklist

-   [x] I have run the linter


MatoPlus:
Hmm I think we can use two things for the prompt here:
1. When we click on the "Yes" button, we should keep the modal open and let the button enter the loading state via `loading={true}` prop. We do not want the prompt to close until the class has actually been unregistered
2. In the description of the modal. I think we should inform them to contact SDC for a refund as we do not handle it automatically. @uwblueprint/sdc-design any thoughts on what the modal should say?
3. We will need translation for any new text in the external platform. Please look at the examples in the `common.json` file and add the new text accordingly

lmk if you need help with any of these steps!

bekahma:
is this a new modal design? it looks slightly different than the existing modal i remember but i could be tripping

PR #177, opened 2021-12-08
external/internal changes showing expired crim checks

aamyli:
### Notion ticket link

[add expiry to criminal checks](https://www.notion.so/uwblueprintexecs/Add-expiry-date-to-criminal-checks-1-year-on-expire-we-should-treat-it-as-not-approved-f3f032d004b44835a6a2cd8afc10e9ff)

added expiry check to both external & internal application; for consistency, the definition for expired is: **if crim check is approved & submittedat is > 1 year**

EXTERNAL APPLICATION:
Screen Shot 2021-12-26 at 11 30 50 PM

Screen Shot 2021-12-26 at 11 31 08 PM
&& also added an info line for approved crim checks notifying they only last one year -- to adjust translations as well if we decide to keep
Screen Shot 2021-12-26 at 11 33 14 PM

INTERNAL APPLICATION:
Screen Shot 2021-12-27 at 12 06 19 AM
Screen Shot 2021-12-27 at 12 06 36 AM

- added logic to check for crim check expiry and corresponding translations  

### Steps to test

1. can see if the document is approved & submitted > 1 year ago, should notify expiry 

-

### Checklist

-   [x] I have run the linter


bekahma:
those changes look good to me :) what are you planning to do with the `account.status`?

aamyli:
> those changes look good to me :) what are you planning to do with the `account.status`?

@bekahma oh i think that's just there because i didn't apply the translations stuff yet haha ignore that; but yeah okay as long as you're good with the copies i'll stick with them! let me know if you want something more descriptive in the myaccounts page (eg indicating that crim checks are only valid for one year)

bekahma:
i think it's ok to leave that part out about expiring after 1 year... if we want to include it i would prefer notifying the user before it's expired (i.e. in the instructions for uploading criminal checks) - what do you think?

public/locales/en/common.json
```diff
@@ -54,13 +54,15 @@
...
         "status_approved": "Status: Approved",
         "status_pending": "Status: Pending",
+        "status_expired": "Status: Expired",
         "status_declined": "Status: Declined",
         "dateSubmitted": "Date submitted: {{date}}",
         "poi_approved": "Your income statement has been approved. Please be on the lookout for an email from SDC with the discount for future registrations.",
         "poi_pending": "Your income statement is under review.",
         "poi_declined": "You have been declined for low income discounts. Please contact SDC if you think this is a mistake or submit a new income statement.",
-        "bgc_approved": "Your criminal record check has been approved.",
+        "bgc_approved": "Your criminal record check has been approved. It is valid for 1 year from date of submission.",
```
aamyli, line 63: to adjust the translations as well if we decide to keep this additional line

pages/admin/registrant/user/[id].tsx
```diff
@@ -173,8 +174,7 @@ export default function Registrant(props: AdminProps): JSX.Element {
                             />
                         ) : (
                             <EmptyState height="200px">
-                                The participant has not uploaded a criminal record check at this
-                                time.
+                                The participant has not uploaded a proof of income at this time.
```
Comment (aamyli) [inline pages/admin/registrant/user/[id].tsx:177]: fixed the wording here for poi

public/locales/ko/common.json
```diff
@@ -47,6 +47,7 @@
         "save": "정보 저장",
         "status_approved": "상태: 승인",
         "status_pending": "상태: 보류 중",
+        "status_expired": "TEMP HOLD",
```
MatoPlus: Don't add a placeholder. It's easier for third-party programs like https://www.electronjs.org/apps/i18n-manager to pick up if the translations are missing. 
```suggestion
```

src/components/MissingDocAlert.tsx
```diff
@@ -60,7 +70,7 @@ export const MissingDocAlert: React.FC<MissingDocAlertProps> = ({ me }) => {
                         <Box height={{ base: "35px", md: "50px", lg: "70px" }}>
                             <InfoIcon />
                         </Box>
-                        <Box w="60%">
+                        <Box w="90%">
```
MatoPlus: Just a note, using percentages like these generally just us results that are not mobile responsive. I think it should be fine in this case but I can test this later

public/locales/zh/common.json
```diff
@@ -47,6 +47,7 @@
         "save": "保存更改",
         "status_approved": "状态：批准",
         "status_pending": "状态：待定",
+        "status_expired": "TEMP HOLD",
```
MatoPlus: idem
```suggestion
```

utils/checkExpiry.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Check if date is expired (defining expired as 1 year from the date)
+ * @param date the date we to check
+ * @returns true if expired, false if not
+ */
+export default function checkExpiry(date: Date): boolean {
+    const d = new Date(date);
+    const expiryDate = new Date(d.getUTCFullYear() + 1, d.getMonth(), d.getDate());
```
MatoPlus: Any reason why you are mixing UTC year with client month/dates? Sine this is only used client side, maybe do:
```suggestion
    const expiryDate = new Date(d.getFullYear() + 1, d.getMonth(), d.getDate());
```

MatoPlus (changes requested):
It's looking good! a couple of nits.

src/components/MissingDocAlert.tsx
aamyli: oh okay, i just changed the width because the formatting felt a bit wonky but i didn't really take mobile into account, i'll put it back to 60!

MatoPlus: oh i meant just using percentages that are not "100%" in general.

MatoPlus: @aamyli You can revert it back to 90% if you think it's appropriate. As long as you've tested how it looks in both web and mobile view, it's good to ship.

MatoPlus (approved):
LGTM!

PR #181, opened 2021-12-10
Fixed Bug where "Update Class" is disabled when some fields take awhile to load

gregmaxin:
### Notion ticket link

-

### Steps to test

1.

-

### Checklist

-   [ ] I have run the linter

### DIFF
--- a/pages/admin/class/edit/[id].tsx
+++ b/pages/admin/class/edit/[id].tsx
@@ -349,7 +349,7 @@ export default function CreateClass({ session }: Props): JSX.Element {
                 ></CheckBoxField>
                 {EDIT ? (
                     <Button
-                        key={className} //When loading from localstorage finishes this causes the button to re-render
+                        key={"Edit Class_" + className + "_" + price + "_" + programs.length}
                         id="Submit"
                         bg={colourTheme.colors.Blue}
                         color={"white"}


pages/admin/class/edit/[id].tsx
```diff
@@ -349,7 +349,7 @@ export default function CreateClass({ session }: Props): JSX.Element {
                 ></CheckBoxField>
                 {EDIT ? (
                     <Button
-                        key={className} //When loading from localstorage finishes this causes the button to re-render
+                        key={"Edit Class_" + className + "_" + price + "_" + programs.length}
```
Comment (MatoPlus) [inline pages/admin/class/edit/[id].tsx:352]: that's a pretty beefy key 😳

MatoPlus:
@GodGreg can you explain how your changes fix the issue?

pages/admin/class/edit/[id].tsx
Comment (MatoPlus) [reply pages/admin/class/edit/[id].tsx:352]: Doesn't look like this fixed the problem. The troublesome field we're looking at here is the Teacher select specifically. I don't think tweaking the key here will help.

image

MatoPlus:
@GodGreg I don't think this is quite right. Do you think we can close this and merge this fix: https://github.com/uwblueprint/social-diversity-for-children/pull/185 ? I think my PR should fix the problem in question

MatoPlus:
merged #185 instead.

PR #182, opened 2021-12-10
fixed waitlist bug

aamyli:
### Notion ticket link

[ticket from bug bash](https://www.notion.so/uwblueprintexecs/Waitlist-behaviour-is-janky-multiple-classes-on-waitlist-52237c8e1cfc49c4840d6925af30b0b8)

- removed `WaitlistInput` and is passing in `classId` and `parentId` everywhere instead for consistency

### Steps to test

1. Try having multiple users on waitlist + play around with removing / adding to waitlist, etc. 

I tested it out with 2 users + a couple different scenarios and everything looks good thus far

-

### Checklist

-   [x] I have run the linter

### DIFF
--- a/pages/api/waitlist/index.ts
+++ b/pages/api/waitlist/index.ts
@@ -22,12 +22,11 @@ export default async function handle(req: NextApiRequest, res: NextApiResponse):
     if (!parentId) {
         return ResponseUtil.returnBadRequest(res, "No user id stored in session");
     }
+    const { classId } = req.body;
 
     switch (req.method) {
         // Retrieving a waitlist by passing in the classId and parentId.
         case "GET": {
-            const { classId } = req.body;
-
             if (classId && parentId) {
                 // If both classId and parentId are passed in, retrieve the specific waitlist record
                 const waitlistRecord = await getWaitlistRecord(classId, parentId);
@@ -50,8 +49,6 @@ export default async function handle(req: NextApiRequest, res: NextApiResponse):
             break;
         }
         case "POST": {
-            const { classId } = req.body;
-
             const validationErrors = validateWaitlistRecord(classId, parentId);
             if (validationErrors.length !== 0) {
                 ResponseUtil.returnBadRequest(res, validationErrors.join(", "));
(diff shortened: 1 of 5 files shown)


pages/api/waitlist/index.ts
```diff
@@ -19,6 +18,8 @@ import { validateWaitlistRecord } from "@utils/validation/waitlist";
 export default async function handle(req: NextApiRequest, res: NextApiResponse): Promise<void> {
     const session = await getSession({ req });
 
+    console.log(session);
+
     const parentId = session.id as number;
```
MatoPlus: I think session.id is always a number, you can refer to `types/next-auth.d.ts`
```suggestion
    const parentId = session.id;
```

pages/api/waitlist/index.ts
```diff
@@ -19,6 +18,8 @@ import { validateWaitlistRecord } from "@utils/validation/waitlist";
 export default async function handle(req: NextApiRequest, res: NextApiResponse): Promise<void> {
     const session = await getSession({ req });
 
+    console.log(session);
+
```
MatoPlus: yeet
```suggestion
```

MatoPlus (changes requested):
oh can we remove the Waitlist type from the model now?

pages/api/waitlist/index.ts
```diff
@@ -27,19 +26,19 @@ export default async function handle(req: NextApiRequest, res: NextApiResponse):
     switch (req.method) {
         // Retrieving a waitlist by passing in the classId and parentId.
         case "GET": {
-            const input = req.body as WaitlistInput;
+            const { classId } = req.body;
```
MatoPlus: Maybe we move this up to remove code redundancy? Seems like it's used in all cases.

MatoPlus (approved):
LGTM! If you've tested in staging and test works with multiple volunteers account, feel free to merge 🚢 

otherwise, I'll test it one more time tmw and merge.

MatoPlus:
I think I found out why the waitlist feature didn't work in staging in the bug blitz. We are using our mailing service in our prod account but our API key is from our test account. This means that waitlist emails would not be sent out. As long as this works locally, this should be good to go!

MatoPlus (approved):
Does not work on staging but works locally with valid email server. LGTM!

PR #185, opened 2021-12-12
fix(edit/class): useEffect depends on user query

MatoPlus:
### Notion ticket link

[Edit Class Teacher Does not load in time](https://www.notion.so/uwblueprintexecs/Edit-Class-Teacher-Does-not-load-in-time-4b7a97ff31914f8bb026798938697d8c)

When loading the edit class page, the teacher field does not populate on the first try when cache is empty. The useEffect is not complete, it needs to depend on the useUsers query.


MatoPlus:
Merging in favour of https://github.com/uwblueprint/social-diversity-for-children/pull/181

PR #187, opened 2021-12-12
fix: 404 page support static translation

MatoPlus:
### Notion ticket link

[Fix 404 page, some wrapper components not translated](https://www.notion.so/uwblueprintexecs/Fix-404-page-some-wrapper-components-not-translated-d3452f1c22f24c9e808f4133231e323e)

image

- Use static props function to support translations as shown here

### Steps to test

1. go to any 404 page with locale set. test with all locales. we should no longer see things like "nav.signInNow" but rather the actual text.

### Checklist

-   [X] I have run the linter

### DIFF
--- a/pages/404.tsx
+++ b/pages/404.tsx
@@ -3,24 +3,23 @@ import SvgErrorIcon from "@components/icons/ErrorIcon";
 import colourTheme from "@styles/colours";
 import { Button, Box, Center, Heading, Text } from "@chakra-ui/react";
 import Link from "next/link";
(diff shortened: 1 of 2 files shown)


feng-eric (approved):
LGTM. How come `ja/common.json` and `ja/form.json` are included but empty?

MatoPlus:
> LGTM. How come `ja/common.json` and `ja/form.json` are included but empty?

SDC does not have JP translators so those files will remain empty for now. i18n is smart enough to fallback to english

PR #194, opened 2021-12-15
add prod terraform

matthewgeng:
### Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/6dd603d3687743ce8d09d6298dc99748?v=7332b6570bcc4b48b5cd73fa2719bec8&p=2c7608b53c6c4781b45b627f6092e33f)

- Add production terraform. Same as staging. Currently the s3 bucket names have prod embedded in the name whereas in staging it does. I.e. prod: `sdc-prod-uploads` vs staging: `sdc-uploads`. This is something we should look into fixing.

### Steps to test

1.

-

### Checklist

-   [x] I have run the linter

### DIFF
--- a/terraform/environments/production/main.tf
+++ b/terraform/environments/production/main.tf
@@ -6,17 +6,17 @@ terraform {
   }
   # not possible to have this dynamic or populated through variables 
   backend "s3" {
-    bucket         = "sdc-prod-terraform-state"
+    bucket         = "sdc-app-terraform-state"
     key            = "terraform-state/terraform.tfstate"
-    region         = "us-east-1"
-    dynamodb_table = "sdc-prod-terraform-state-lock"
+    region         = "us-west-2"
+    dynamodb_table = "sdc-app-terraform-state-lock"
     encrypt        = true
   }
 }
 
 # configures the required provider
 provider "aws" {
-  region = "us-east-1"
+  region = "us-west-2"
 }
 
 module "iam" {
@@ -29,11 +29,10 @@ module "s3" {
   allowed_origins = [var.sdc_domain, var.sdc_pr_domain]
 
   # uploads bucket 
-  s3_uploads_bucket_name  = var.s3_uploads_bucket_name
-  criminal_check_folder   = var.criminal_check_folder
-  income_proof_folder     = var.income_proof_folder
-  curriculum_plans_folder = var.curriculum_plans_folder
-  other_folder            = var.other_folder
+  s3_uploads_bucket_name = var.s3_uploads_bucket_name
+  criminal_check_folder  = var.criminal_check_folder
+  income_proof_folder    = var.income_proof_folder
+  other_folder           = var.other_folder
 
   # images bucket
   s3_images_bucket_name = var.s3_images_bucket_name
--- a/terraform/environments/production/main.tf
+++ b/terraform/environments/production/main.tf
@@ -73,6 +73,6 @@ module "cronMailing_eventbridge" {
   source              = "../../modules/eventbridge"
   rule_name           = var.cronMailing_rule_name
   schedule_expression = var.cronMailing_schedule_expression
-  target_arn          = module.cronMailing.lamba_function_arn
-  target_id           = module.cronMailing.lamba_function_name
+  target_arn          = module.cronMailing.lambda_function_arn
+  target_id           = module.cronMailing.lambda_function_name
 }
--- a/terraform/environments/production/main.tf
+++ b/terraform/environments/production/main.tf
@@ -26,7 +26,7 @@ module "iam" {
 
 module "s3" {
   source          = "../../modules/s3"
-  allowed_origins = [var.sdc_domain, var.sdc_pr_domain]
+  allowed_origins = [var.sdc_domain]
 
   # uploads bucket 
   s3_uploads_bucket_name = var.s3_uploads_bucket_name
--- a/terraform/environments/production/main.tf
+++ b/terraform/environments/production/main.tf
@@ -1,0 +1,79 @@
+terraform {
+  required_providers {
+    aws = {
+      source = "hashicorp/aws"
+    }
+  }
+  # not possible to have this dynamic or populated through variables 
+  backend "s3" {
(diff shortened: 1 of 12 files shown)


terraform/environments/production/variables.tf
```diff
@@ -0,0 +1,123 @@
...
+  description = "Name of the income proof folder"
+  default     = "income-proof"
+  type        = string
+}
+
+variable "curriculum_plans_folder" {
+  description = "Name of the curriculum plans folder"
+  default     = "curriculum-plans"
+  type        = string
+}
```
MatoPlus: we technically don't need this folder any more

terraform/environments/production/variables.tf
```diff
@@ -0,0 +1,123 @@
...
+  default     = "cloudwatch-lambda-logs-policy"
+  type        = string
+}
+
+# ------------------------------------------------------------------
+# s3
+
+variable "s3_images_bucket_name" {
+  description = "Name of the document uploads s3 bucket"
+  default     = "sdc-prod-public-images"
```
MatoPlus: I was thinking maybe instead of using prod to differentiate s3 folder names, we can use something like `sdc-app-public-images`. Thoughts?

terraform/environments/production/variables.tf
```diff
@@ -0,0 +1,123 @@
...
+  type        = string
+}
+
+variable "s3_uploads_bucket_name" {
+  description = "Name of the document uploads s3 bucket"
+  default     = "sdc-prod-uploads"
```
MatoPlus: idem

terraform/environments/production/remote_state/main.tf
```diff
@@ -0,0 +1,55 @@
+terraform {
+  required_providers {
+    aws = {
+      source = "hashicorp/aws"
+    }
+  }
+}
+
+provider "aws" {
+  region = "us-east-1"
```
MatoPlus: we actually want to run prod in us-west-2 or us-west-1. From my understanding, we can probably choose the one with less demand cost, (us-west-2)?

terraform/environments/production/main.tf
```diff
@@ -0,0 +1,79 @@
...
+  required_providers {
+    aws = {
+      source = "hashicorp/aws"
+    }
+  }
+  # not possible to have this dynamic or populated through variables 
+  backend "s3" {
+    bucket         = "sdc-prod-terraform-state"
+    key            = "terraform-state/terraform.tfstate"
+    region         = "us-east-1"
```
MatoPlus: us-west-2 or 1 is closer to our client!

terraform/environments/production/main.tf
```diff
@@ -0,0 +1,79 @@
...
+    dynamodb_table = "sdc-prod-terraform-state-lock"
+    encrypt        = true
+  }
+}
+
+# configures the required provider
+provider "aws" {
+  region = "us-east-1"
```
MatoPlus: idem

MatoPlus (changes requested):

terraform/environments/production/variables.tf
matthewgeng: ah okay, I'll also remove from dev then?

terraform/environments/production/variables.tf
matthewgeng: ah okay yeah I'm good with app!

terraform/environments/production/remote_state/main.tf
matthewgeng: yeah us-west-2 is cheaper, I'll change to us-west-2

terraform/environments/production/variables.tf
```diff
@@ -0,0 +1,117 @@
...
+  description = "full domain name"
+  default     = "https://app.socialdiversity.org"
+  type        = string
+}
+
+variable "sdc_pr_domain" {
+  description = "domain of pr deploys"
+  default     = "https://social-diversity-for-children-pr-*.up.railway.app"
+  type        = string
+}
```
MatoPlus: I don't think we want to allow PR to access production stuff.
```suggestion
```

MatoPlus (changes requested):
one more change and I think this should be good! ^^ listed above

terraform/environments/production/variables.tf
matthewgeng: removed!

MatoPlus (approved):

terraform/environments/production/main.tf
```diff
@@ -0,0 +1,78 @@
...
+  }
+}
+
+# cron job for lambda cronMailing
+module "cronMailing_eventbridge" {
+  source              = "../../modules/eventbridge"
+  rule_name           = var.cronMailing_rule_name
+  schedule_expression = var.cronMailing_schedule_expression
+  target_arn          = module.cronMailing.lamba_function_arn
+  target_id           = module.cronMailing.lamba_function_name
```
MatoPlus: are these typos btw?
```suggestion
  target_arn          = module.cronMailing.lambda_function_arn
  target_id           = module.cronMailing.lambda_function_name
```

terraform/environments/production/variables.tf
```diff
@@ -0,0 +1,111 @@
...
+# environment variables
+
+variable "api_endpoint" {
+  description = "Domain api endpoint"
+  default     = "https://app.socialdiversity.org/api"
+  type        = string
+}
+
+variable "lambda_secret_key_name" {
+  description = "Name of lambda secret key"
```
MatoPlus, line 109: I think we can have a default value here right? I don't think we need to provide the secret value every time we try are trying to plan.

matthewgeng: Yeah don't mind a default value here! I'll add that and remove the secret in github

terraform/environments/production/main.tf
matthewgeng: yup fixed!

matthewgeng: Good catch!

PR #195, opened 2021-12-17
fix(api): update document resets approved

MatoPlus:
### Notion ticket link

[BUG: when submitting new criminal check or income proof, switch current approval status to false as precaution ](https://www.notion.so/uwblueprintexecs/BUG-when-submitting-new-criminal-check-or-income-proof-switch-current-approval-status-to-false-as--b1e7b472f4bc485799ad9cfc7cb717b0)

- When updating bgc or poi of user, also reset the approved status to null
- fixed default value of is low income to null, which is pending state

### Steps to test

1. As volunteer, submit bgc, approve it in the DB, and submit another doc. the second doc should be pending
2. same as step one but for parent poi

### Checklist

-   [X] I have run the linter

### DIFF
--- a/prisma/schema.sql
+++ b/prisma/schema.sql
@@ -89,7 +89,7 @@ CREATE TABLE classes (
 CREATE TABLE parents (
   id SERIAL PRIMARY KEY NOT NULL,
   phone_number VARCHAR(50) NOT NULL,
-  is_low_income BOOLEAN DEFAULT false,
+  is_low_income BOOLEAN,
   preferred_language locales NOT NULL,
   proof_of_income_link TEXT,
   proof_of_income_submitted_at TIMESTAMPTZ,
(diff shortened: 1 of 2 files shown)


prisma/schema.sql
```diff
@@ -89,7 +89,7 @@ CREATE TABLE classes (
 CREATE TABLE parents (
   id SERIAL PRIMARY KEY NOT NULL,
   phone_number VARCHAR(50) NOT NULL,
-  is_low_income BOOLEAN DEFAULT false,
+  is_low_income BOOLEAN,
```
feng-eric, line 92: @MatoPlus Can you explain why false isn't equivalent to the "pending" state?

MatoPlus: My understanding is that we have three states, denied, pending, and approved. This reflects our design for internal but not external (atm since we only have pending and approved for external).

feng-eric (approved):
LGTM

PR #197, opened 2021-12-18
chore(api): add proper RBAC for api to reduce access

MatoPlus:
### Notion ticket link

[Patch All API Endpoints for minimal access - RBAC](https://www.notion.so/uwblueprintexecs/Patch-All-API-Endpoints-for-minimal-access-RBAC-5082191ef43149baa19d92f2f82f9b8f)

- For all API points, we are ensuring that RBAC checks are performed when appropriate

### Steps to test

1. All frontend pages using affected API does not break if used as intended

- Any missing RBACs
- added RBAC logic implementation

### Checklist

-   [X] I have run the linter

### DIFF
--- a/pages/api/class/index.ts
+++ b/pages/api/class/index.ts
@@ -5,6 +5,8 @@ import { ClassInput, ClassTranslationInput } from "@models/Class";
 import { validateClassData } from "@utils/validation/class";
 import { getClassInfoWithProgramId } from "@database/program-card-info";
 import { stripe } from "services/stripe";
+import { getSession } from "next-auth/client";
+import { isAdmin } from "@utils/session/authorization";
 
 /**
  * handle controls the request made to the class resource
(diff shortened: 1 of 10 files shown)


pages/api/class/index.ts
```diff
@@ -5,21 +5,29 @@ import { ClassInput, ClassTranslationInput } from "@models/Class";
...
  * @param req API request object
  * @param res API response object
  */
 export default async function handle(req: NextApiRequest, res: NextApiResponse): Promise<void> {
+    const session = await getSession({ req });
     switch (req.method) {
         case "GET": {
             const { id: programId, archived } = req.query;
 
+            const archivedParsed = Boolean(JSON.parse((archived as string) || "false"));
```
feng-eric, line 22: @MatoPlus Can you explain this part to me? How come we're doing this before we check `!isAdmin(session)`?

pages/api/user/index.ts
```diff
@@ -73,13 +74,8 @@ export default async function handle(req: NextApiRequest, res: NextApiResponse):
             ResponseUtil.returnOK(res, updatedUser);
             break;
         }
-        case "DELETE": {
-            // TODO:
-            break;
-        }
         default: {
             const allowedHeaders: string[] = ["GET", "POST", "PUT", "DELETE"];
```
feng-eric:
```suggestion
            const allowedHeaders: string[] = ["GET", "POST", "PUT"];
```

pages/api/class/index.ts
MatoPlus: My understanding is that the route is open to the public, but archived items requires admin access. 

quoting the comment: "get archived should be admin only"

pages/api/user/index.ts
MatoPlus: good catch : )

pages/api/class/index.ts
feng-eric: Might be good to leave a comment here or perhaps make sure this is documented somewhere along with which other routes should be private

feng-eric (approved):

pages/api/class/index.ts
MatoPlus: i'll make the comment a bit more intuitive, we're just following the least privilege access principle for our api routes.

PR #207, opened 2021-12-20
Release/v0.0.1

matthewgeng:
Release Log: https://www.notion.so/uwblueprintexecs/Release-Log-e6d5dc2a0519409dab786323d2909eee

Second attempt due to bad history


MatoPlus:
Tested on PR deploy, only issue with deployment is that the region of the SES server is in the wrong region. Not blocking, will delegate for next release.

MatoPlus (approved):

PR #209, opened 2021-12-21
chore(env): add missing sample env

MatoPlus:
Add missing file in env sample so build will fail and pick up on missing variables


.env.sample
```diff
@@ -9,4 +9,5 @@ AWS_REG=
 AWS_LAMBDA_SECRET=
 STRIPE_SECRET_KEY=
 NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY=
-S3_UPLOAD_BUCKET=
\ No newline at end of file
+S3_UPLOAD_BUCKET=
+S3_PUBLIC_IMAGES_BUCKET=
```
MatoPlus, line 13:
```suggestion
S3_PUBLIC_IMAGES_BUCKET=

```

MatoPlus:
dup

PR #220, opened 2021-12-29
feat(db): init prod migrations

MatoPlus:
### Notion ticket link

[Setup-Simple-Migration-Workflow-For-Production-Dev](https://www.notion.so/uwblueprintexecs/Setup-Simple-Migration-Workflow-For-Production-Dev-d1428b72873d4fc6a8f9f121aeffac2f)

- Add manual migration flow as stated in changes of readme file
- Removed unnecessary schema folder
- Init migration and simple migration to remove deprecated field

Here is what I did for our init migration:

```bash
# This created our init migration sql file. This required me to reset the staging or local DB as migration history is initially not in order. For production, we can simply baseline our migrations to avoid resets as stated in README
npx prisma migrate dev --name init --create-only 

# Change `schema.sql` and `schema.prisma` to remove starttime_minutes and run the following:
npx prisma migrate dev --name remove_starttime_minutes_field --create-only   # this generated the second migration sql file

# we can now deploy our migrations to our DB
npx prisma migrate deploy 
```

NOTE: We do not need a env for shadow DB in prod according to: https://www.prisma.io/docs/concepts/components/prisma-migrate/shadow-database since it's only ever used for local commands.

- Flow of migration?
- Next steps - automate with CI @uwblueprint/sdc-leads @uwblueprint/sdc-infra  open to thoughts!

### Checklist

-   [X] I have run the linter


MatoPlus:
For our CI. Here is my thoughts on automating this.

Two options:
1. Just make this a manual process. Everytime we want to do a migration. We manually try to do a `npx prisma migrate deploy` and if there are any issues, we manually take care of any issues with `prisma migrate resolve --applied <MIGRATION_FOLDER_NAME>`
2. On push to main, we do a `npx prisma migrate deploy` we either get told that there are no pending migrations or that the migration has failed. If the migration fails, we want to shut down railway app and manually debug the problem via `migrate resolve`. Ideallly, in the release PR, we do a `npx prisma migrate status` and `npx prisma db pull` to ensure that the prod DB is in the expected state for the migration (ie: it is missing the expected migration files and the schema pulled does not have any oddities outside of the migration patch).

Also, migrations should apply to staging when possible. ie: on push to staging, we should have the same workflow.

@uwblueprint/sdc-infra thoughts?

MatoPlus:
After this PR gets merged in. I would like to do another release ASAP. It does not look like our temp prod DB is used atm so this is a great chance to get our initial migration history in sync. we could resolve the init migration and just apply the migration to remove the deprecated field but it's safer to just reset and apply all migrations from scratch since our init migration does not match the current prod schema (isLowIncome should not be default False).

MatoPlus:
Merging this soon to unblock other incoming PR changes that require schema tweaks. Please ping @uwblueprint/sdc-infra or me if there are any concerns.

matthewgeng:
Whoops! I didn’t submit my review my bad! But yeah looked good to me!

MatoPlus:
Just to throw the update here. The migration scripts were applied successfully on staging. The following steps were executed after setting the right values in the `.env` file:

```bash
heroku pg:reset -a <staging-app>
npx prisma migrate deploy
```

when we merge in production, we would like to run the same thing for the first time (ie: this will reset all the data for the first time). In the future, we would no longer need to reset the DB after syncing the migration history initially.

PR #221, opened 2022-01-02
fix: cardInfoUtil startTime convert

MatoPlus:

- we want explicit date conversion when converting API results into Date objects

### Steps to test

1. Program Details Page actually loads properly without client error


utils/cardInfoUtil.ts
```diff
@@ -48,7 +48,7 @@ export class CardInfoUtil {
             startDate: new Date(result.startDate),
             endDate: new Date(result.endDate),
             weekday: result.weekday,
-            startTimeMinutes: totalMinutes(result.startDate),
+            startTimeMinutes: result.startDate ? totalMinutes(new Date(result.startDate)) : 0,
```
feng-eric, line 51: Pretty minor but on line 48, you already have `new Date(result.startDate)`. Up to you if you want to reduce duplicate code.

feng-eric: Also, in the cases when the ternary expression is false, what would `new Date(result.startDate)` return on line 48? Should there be a ternary expression there too then?

feng-eric (dismissed):

utils/cardInfoUtil.ts
MatoPlus: true, I'll do new Date() in for false I guess.

MatoPlus: I think I'll just leave the one dup line in for now. I think it should be fine with how we're doing ternaries.

feng-eric (approved):

PR #231, opened 2022-01-06
feat(translations): add zh translations

MatoPlus:


public/locales/en/common.json
```diff
@@ -120,7 +120,7 @@
     "nav": {
         "signInNow": "Sign in now to register for classes",
         "register": "Register",
-        "registerNow": "Register now",
+        "registerNow": "Register Now",
```
MatoPlus, line 123: this change was also requested by client

PR #236, opened 2022-01-09
feat(translations): add new ko labels

MatoPlus:
added translations from client


MatoPlus:
we are still missing two more label translations. Will wait until end of week

feng-eric (approved):

MatoPlus:
Still missing translations. Will merge this for now. Opening a PR for the second part when client is done.

PR #238, opened 2022-01-11
fix(stripe): fix internal error + bad redirect

MatoPlus:
- fix issue where API calls throws 500 if trying to get undefined coupon
- fix issue where mutate causes page to rerender to try to checkout again (which we don't want)


utils/createClassRegistration.ts
```diff
@@ -26,7 +26,5 @@ export async function createClassRegistration(
 
     const response = await fetch("/api/enroll/child", request);
 
-    mutate("/api/enroll/child");
```
feng-eric, line 29: `import { mutate } from "swr";` can be removed also then

feng-eric (dismissed):

